PR #2, opened 2015-08-13
Packaging and More Unit Tests

blink1073:
Use TypeScript for source, distribute a CommonJS module and a typing declaration file.
Use npm for scripting, and mocha for tests.
Use mock Websocket and XMLHttpRequests in tests, which can be run in node or after a browserify.


blink1073:
The Mock Websocket does not work properly on Firefox and Safari, so no Travis karma testing...

blink1073:
Also, we can't test on PhantomJS because it lacks a WeakMap

blink1073:
screen shot 2015-08-13 at 4 56 28 pm

rgbkrk:
Woohoo!

The `lib/` and `dist/` folders shouldn't be part of source control (`.gitignore`'d but not `.npmignore`'d).

How is phosphor being included? Via bower?

blink1073:
This way, someone consuming the package won't need to download the long build chain tail.  Yes, I'm grabbing phosphor with bower.

blink1073:
Forget that last commit, I take it the lib/ folder would be consumed by the npm upload?  If so, I could be convinced either way.

rgbkrk:
Yeah, make a `prepublish` target in your package.json:

```json
  "build": "gulp",
  "prepublish": "npm run build",
```

Only source in source control, `lib/` goes out to npm and `dist` to bower.

By the way, we intended `dist` for bower but haven't yet released anything to bower (since we were continually consuming from npm).

(Worth reading http://blog.keithcirkel.co.uk/how-to-use-npm-as-a-build-tool/ for this as well)

package.json
```diff
@@ -2,43 +2,49 @@
...
+    "notebook"
   ],
   "author": "Project Jupyter",
   "license": "BSD-3-Clause",
   "bugs": {
     "url": "https://github.com/jupyter/jupyter-js-services/issues"
   },
+  "main": "lib/index.js",
   "homepage": "https://github.com/jupyter/jupyter-js-services"
 }
```
jdfreder: I think you want a Typescript block here, pointing to your main d.ts file generated by dts-bundle.  See https://github.com/basarat/ts-npm-module/blob/master/package.json#L6-L8 .

blink1073: Done, thanks!

rgbkrk:
This packaging is definitely looking much better here!

Have you tried out [ws](https://github.com/websockets/ws) for node compatible testing (only as a dev dependency)? It doesn't look like `window` or `document` is used here, so you won't need JSDom.

blink1073:
Thanks Kyle, ws looks very plausible.

minrk:
I use ws for tests [in configurable-http-proxy](https://github.com/jupyter/configurable-http-proxy) if that's a helpful reference.

blink1073:
Thanks Min, that helps with the testing server side.  I'm still not clear on how to handle the client side.  Any thoughts on that?

minrk:
Do you mean in the browser, or using node as the client?

blink1073:
I mean that if I use `ws` in `kernel.ts`, will it play well when I browserify it?

blink1073:
I was going to test it out on Monday.

src/kernel.ts
```diff
@@ -610,9 +607,15 @@ class Kernel {
...
     var ws_host_url = this._wsUrl + this._kernelUrl;
 
-    kernel_log.info("Starting WebSockets:", ws_host_url);
+    console.info("Starting WebSockets:", ws_host_url);
 
-    this._ws = new WebSocket(this.wsUrl);
+    if (typeof(WebSocket) !== 'undefined') {
+      this._ws = new WebSocket(this.wsUrl);
+    } else if (typeof(MozWebSocket) !== 'undefined') {
+      this._ws = MozWebSocket(this.wsUrl);
```
rgbkrk: Which browsers are still using `MozWebSocket`?

blink1073: It was in the old code, and I assumed it was used by Firefox, but the console in 38 says otherwise...

rgbkrk:
What would you think about passing `WebSocket` to the Kernel constructor (to later be referenced as `this.WebSocket` or similar)? This would make testing work more easily across contexts as well as be used in node/iojs applications.

In a similar vein, @jdfreder and I made the output area and transformime accept a `Document` on purpose to work across `iframe`s, vanilla `window.document`, and `jsdom`.

blink1073:
@rgbkrk, it was my understanding that his part would only be used in a browser or Electron, where one would have a stand Websocket and XMLHttpRequest.

blink1073:
for testing, I am just overriding the global WebSocket and XMLHttpRequest variables.

rgbkrk:
That's fine. No decree on that one, primarily a suggestion for modularity. That's why I asked what you thought about it, as part of the interface.

`XHR` is available in `jsdom` as well, off the `window` (which comes from `document.defaultView`).

```
> var document = require('jsdom').jsdom();
> var window = document.defaultView;
> var XMLHttpRequest = window.XMLHttpRequest;
```

blink1073:
Right now XMLHttpRequest is only directly used in `utils.XMLHttpRequest`.  There's nothing stopping someone from declaring `global.XMLHttpRequest = window.XMLHttpRequest`.

blink1073:
I guess my point is, I see those as environmental variables, not as part of the API of this library.

rgbkrk:
You're definitely right, there's nothing stopping anyone from doing that.

rgbkrk:
Totally didn't mean to distract from the excellent packaging work here. Thank you @blink1073! :smile:

blink1073:
Ha!  Thanks for keeping me honest.

blink1073:
The node refactor is almost done, I just have to finish hooking my mock objects up to the existing tests.

blink1073:
I think this one is ready for merging.

minrk:
huzzah!

rgbkrk:
:tada:

PR #9, opened 2015-08-19
Clean Up

blink1073:
Clean up style, reorganize tests.
Fix scoping errors.

### DIFF
--- a/test/src/mocksocket.ts
+++ b/test/src/mocksocket.ts
@@ -19,9 +19,9 @@ declare var global: any;
 class SocketBase {
 
   static CONNECTING = 0;  // The connection is not yet open.
-  static OPEN = 1;  //  The connection is open and ready to communicate.
-  static CLOSING = 2;  //  The connection is in the process of closing.
-  static CLOSED = 3; //  The connection is closed or couldn't be opened.
+  static OPEN = 1;        // The connection is open and ready to communicate.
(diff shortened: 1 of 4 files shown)


test/src/mocksocket.ts
```diff
@@ -82,21 +77,21 @@ class SocketBase {
...
-    if (this._onClose) {
-      utils.doLater(() => {
-        this._readyState = SocketBase.CLOSED;
-        this._onClose();
-      });
-    }
+    utils.doLater(() => {
+      this._readyState = SocketBase.CLOSED;
+      var onClose = this._onClose;
+      if (onClose) onClose();
```
rgbkrk: This one has different logic than before, changing the `_readyState` whether or not `this._onClose` is defined.

rgbkrk: (Just a note - I have no idea what the right behavior is here)

rgbkrk:
These are all for the mocks, so it's all fine to me.

I'm also ok with picking a style and sticking to it with either an eslintrc, jshintrc, or even one of the `*standard*` variants. I'd open that up on the mailing list, but it's not my can of worms.

test/src/mocksocket.ts
blink1073: It is correct in this PR.

PR #11, opened 2015-08-20
Run the docs on Travis so we can catch errors on PRs

blink1073:


rgbkrk:
Make sure to create a branch instead of using master and rebase instead of merging for the future.

blink1073:
Whoops, I'll just incorporate this fix into the next PR.

PR #20, opened 2015-08-25
Overhaul Kernel and Session

blink1073:
- Simplify the lifecycle handling of Kernel and Session -> use factory functions returning Promises.
- Upgrade to TypeScript 1.6.
- Add integration testing with a running notebook server (on Travis as well).
- API example [docs](https://github.com/blink1073/jupyter-js-services/blob/promises_promises/README.md#usage-examples).
- Note: I could not test the `shutdown` or `renameNotebook` methods against the live Notebook server due to cross-origin problems.  The `renameNotebook` issue is that an OPTIONS request is automatically made prior to a cross-origin PATCH request.
  
```
XMLHttpRequest cannot load http://localhost:8888/api/kernels/688730ea-05a3-42af-8a10-df5f4d21d73e. Method DELETE is not allowed by Access-Control-Allow-Methods.
XMLHttpRequest cannot load http://localhost:8888/api/sessions/8c5ee67a-c704-41bf-9c42-b4c68d938c0d. Response for preflight has invalid HTTP status code 405
```
- Expect a follow-on PR to clean up Content and Config and add them to integration tests.
- Coverage:

screen shot 2015-09-09 at 3 42 12 pm


blink1073:
Ready for review.

README.md
```diff
@@ -72,3 +72,96 @@ Earlier versions may also work, but come with no guarantees.
...
+```typescript
+import {
+  IKernelFuture, IKernelId, IKernelInfo, IKernelMsg, Kernel
+} from 'jupyter-js-services';
+
+// get a list of available sessions
+var listPromise = Kernel.list('http://localhost:8000');
+
+// start a specific session
+var kernel = NotebookSession('http://localhost:8000');
```
sccolbert: should this be `new NotebookSession`?

README.md
```diff
@@ -72,3 +72,96 @@ Earlier versions may also work, but come with no guarantees.
...
+listPromise.then((kernelList: IKernelId[]) => {
```
sccolbert: You don't need to explicitly type the `kernelList` arg. The promise carries the type, so the function will be contextually-typed automatically by the compiler. 

```typescript
listPromise.then(kernelList => {
  kernel.start(kernelList[0]);
});
```

README.md
```diff
@@ -72,3 +72,96 @@ Earlier versions may also work, but come with no guarantees.
...
+  kernel.start(kernelList[0]);
+});
+
+// send a message and get replies
+kernel.onReady.then(() => {
+  var future: IKernelFuture = kernel.execute('a = 1');
```
sccolbert: You don't need to type `future` here; it can be inferred from `kernel.execute`.

README.md
```diff
@@ -72,3 +72,96 @@ Earlier versions may also work, but come with no guarantees.
...
+  future.onDone(() => {
+    console.log('Future is fulfilled');
+  });
+  future.onOutput((msg: IKernelMsg) => {
```
sccolbert: `msg` can be contextually typed.

README.md
```diff
@@ -72,3 +72,96 @@ Earlier versions may also work, but come with no guarantees.
...
+    console.log(msg.content);  // rich output data
+  });
+});
+
+// restart the kernel and send an inspect message
+kernel.restart().then(() => {
+  var future = kernel.inspect('hello', 5);
+  future.onReply((msg: IKernelMsg) => {
```
sccolbert: `msg` can be contextually typed

README.md
```diff
@@ -72,3 +72,96 @@ Earlier versions may also work, but come with no guarantees.
...
+    console.log(msg.content);
+  });
+});
+
+// interrupt the kernel and send a complete message
+kernel.interrupt().then(() => {
+  var future = kernel.complete('impor', 4);
+  future.onReply((msg: IKernelMsg) => {
```
sccolbert: `msg` can be contextually typed.

README.md
```diff
@@ -72,3 +72,96 @@ Earlier versions may also work, but come with no guarantees.
...
+  });
+});
+
+// register a callback for when the kernel changes state
+kernel.statusChanged.connect((status: string) => {
+  console.log('status', status);
+});
+
+// kill the kernel
+kernel.delete();
```
sccolbert: do we want to rename this to `dispose()`?

README.md
```diff
@@ -72,3 +72,96 @@ Earlier versions may also work, but come with no guarantees.
...
+// interrupt the kernel and send a complete message
+kernel.interrupt().then(() => {
+  var future = kernel.complete('impor', 4);
+  future.onReply((msg: IKernelMsg) => {
+    console.log(msg.content);
+  });
+});
+
+// register a callback for when the kernel changes state
+kernel.statusChanged.connect((status: string) => {
```
sccolbert: `status` can be contextually typed.

README.md
```diff
@@ -72,3 +72,96 @@ Earlier versions may also work, but come with no guarantees.
...
+import {
+  ISessionId, ISessionOptions, NotebookSession
+} from 'jupyter-js-services';
+
+// get a list of available sessions
+var listPromise = NotebookSession.list('http://localhost:8000');
+
+// start a specific session
+var session = NotebookSession('http://localhost:8000');
+listPromise.then((sessionList: ISessionId[]) => {
```
sccolbert: `sessionList` can be contextually typed.

README.md
```diff
@@ -72,3 +72,96 @@ Earlier versions may also work, but come with no guarantees.
...
+  session.start(sessionList[0]);
+});
+
+// restart a session and send a complete message to the kernel
+session.restart().then(() => {
+  var future = session.kernel.complete('impor', 4);
+  future.onReply((msg: IKernelMsg) => {
```
sccolbert: `msg` can be contextually typed

README.md
```diff
@@ -72,3 +72,96 @@ Earlier versions may also work, but come with no guarantees.
...
+    console.log(msg.content);
+  });
+});
+
+// rename the notebook
+session.renameNotbook('/path/to/notebook.ipynb');
+
+// register a callback for when the session changes state
+session.statusChanged.connect((status: string) => {
```
sccolbert: `status` can be contextually typed.

README.md
```diff
@@ -72,3 +72,96 @@ Earlier versions may also work, but come with no guarantees.
...
+  console.log('status', status);
+});
+
+// kill the session
+session.delete();
```
sccolbert: do we want to rename this to `dispose()`?

README.md
```diff
@@ -72,3 +72,96 @@ Earlier versions may also work, but come with no guarantees.
...
+omit the type declarations when using a language other than TypeScript.
+
+**Kernel**
+
+```typescript
+import {
+  IKernelFuture, IKernelId, IKernelInfo, IKernelMsg, Kernel
+} from 'jupyter-js-services';
+
+// get a list of available sessions
```
sccolbert: s/sessions/kernels

README.md
```diff
@@ -72,3 +72,96 @@ Earlier versions may also work, but come with no guarantees.
...
+var listPromise = Kernel.list('http://localhost:8000');
+
+// start a specific session
```
sccolbert: s/session/kernel

README.md
```diff
@@ -72,3 +72,96 @@ Earlier versions may also work, but come with no guarantees.
...
+```typescript
+import {
+  ISessionId, ISessionOptions, NotebookSession
+} from 'jupyter-js-services';
+
+// get a list of available sessions
+var listPromise = NotebookSession.list('http://localhost:8000');
+
+// start a specific session
+var session = NotebookSession('http://localhost:8000');
```
sccolbert: Should this be `new NotebookSession(...)`?

README.md
```diff
@@ -72,3 +72,96 @@ Earlier versions may also work, but come with no guarantees.
...
+// restart a session and send a complete message to the kernel
+session.restart().then(() => {
+  var future = session.kernel.complete('impor', 4);
+  future.onReply((msg: IKernelMsg) => {
+    console.log(msg.content);
+  });
+});
+
+// rename the notebook
+session.renameNotbook('/path/to/notebook.ipynb');
```
sccolbert: s/renameNotbook/renameNotebook

README.md
```diff
@@ -72,3 +72,96 @@ Earlier versions may also work, but come with no guarantees.
...
+
+// register a callback for when the session changes state
+session.statusChanged.connect((status: string) => {
+  console.log('status', status);
+});
+
+// kill the session
+session.delete();
+
+```
```
sccolbert: Run the usage examples through the compiler to make sure they compile cleanly.

src/kernel.ts
```diff
@@ -157,7 +151,7 @@ interface IKernelFuture extends IDisposable {
  * used to communicate with the kernel.
  */
 export
-class Kernel {
+class Kernel extends utils.ReadyPromise {
```
sccolbert: Instead of inheriting from `ReadPromise`, make `ReadPromise` an object which is composed by the consumer. That will allow you to make the protected methods of `ReadyPromise` public, without cluttering up the API of the `Kernel` object. `Kernel` **has a** `ReadyPromise`, not *is a* `ReadyPromise`.

src/kernel.ts
```diff
@@ -355,16 +356,19 @@ class Kernel {
...
    */
-  start(id?: IKernelId): Promise<IKernelId> {
+  start(id?: IKernelId): Promise<IKernelInfo> {
+    this._createReadyPromise();
     if (id !== void 0) {
       this.id = id.id;
       this.name = id.name;
     }
     if (!this._kernelUrl) {
+      console.log('***throwing the error');
```
sccolbert: remove debug log

src/session.ts
```diff
@@ -51,7 +51,7 @@ interface ISessionOptions {
  * all other operations, the kernel object should be used.
  **/
 export
-class NotebookSession {
+class NotebookSession extends utils.ReadyPromise {
```
sccolbert: Same advice r.e. composition vs inheritance

README.md
rgbkrk: I'm a fan of `dispose` as well.

src/kernel.ts
rgbkrk: :+1: to composition here.

blink1073:
All comments addressed.

src/kernel.ts
```diff
@@ -328,12 +330,16 @@ class Kernel {
    *
    * Restart the kernel.
    */
-  restart(): Promise<IKernelId> {
+  restart(): Promise<IKernelInfo> {
+    if (this._promiseDelegate.settled) {
+      console.log("***created a new delegate");
```
sccolbert: remove debug log

src/kernel.ts
```diff
@@ -342,10 +348,12 @@ class Kernel {
       }
       validateKernelId(success.data);
       this.connect();
-      return success.data;
     }, (error: utils.IAjaxError) => {
       this._onError(error);
+    }).catch((data?: any) => {
+      this._promiseDelegate.reject(data);
```
sccolbert: What if the current promise is not settled, but is pending some other operation? The old one gets blow away, but the pending request operates on the new delegate.

src/kernel.ts
```diff
@@ -364,7 +372,10 @@ class Kernel {
       throw Error('You must set the kernel id before starting.');
     }
     this._handleStatus('starting');
-    return utils.ajaxRequest(this._kernelUrl, {
+    if (this._promiseDelegate.settled) {
```
sccolbert: Same comment, having to check whether the promise is settled doesn't seem right for some reason. I want to make sure we have a full understanding of what's going on, and the potential issues this could cause in the present of multiple in-flight requests.

src/kernel.ts
```diff
@@ -373,20 +384,22 @@ class Kernel {
...
    *
-   * Shut down a kernel. Note: if useing a session, Session.shutdown()
+   * Shut down a kernel. Note: if using a session, Session.dispose()
    * should be used instead.
    */
-  shutdown(): Promise<void> {
+  dispose(): Promise<void> {
     this._handleStatus('shutdown');
     this.disconnect();
     return utils.ajaxRequest(this._kernelUrl, {
```
sccolbert: Erm, I think `shutdown` was named fine. It was specifically `delete` I felt was named wrong. If `delete` never actually existed, then this is a non-issue.

I would not expect `dispose` to return a value, so actually prefer `shutdown`.

src/kernel.ts
```diff
@@ -435,13 +449,14 @@ class Kernel {
    * standard HTTP request, but useful function nonetheless for
    * reconnecting to the kernel if the connection is somehow lost.
    */
-  reconnect(): void {
+  reconnect(): Promise<IKernelInfo> {
     if (this.isConnected) {
       return;
```
sccolbert: should this return `onReady` instead of `undefined`?

src/session.ts
```diff
@@ -159,7 +176,7 @@ class NotebookSession {
    *
    * Kill the kernel and shutdown the session.
    */
-  delete(): Promise<void> {
+  dispose(): Promise<void> {
```
sccolbert: I think this should probably be `shutdown` just like the kernel. I don't like `dipose` returning a value, because that is contrary to the `IDisposable` interface.

src/utils.ts
```diff
@@ -159,3 +159,60 @@ function ajaxRequest(url: string, settings: IAjaxSetttings): Promise<any> {
...
+   * Get the underlying Promise.
+   */
+  get promise(): Promise<T> {
+    return this._promise;
+  }
+
+  /**
+   * Test whether the Promise has been settled.
+   */
+  get settled(): boolean {
```
sccolbert: Not sure I like exposing whether or not it's been settled.

src/utils.ts
```diff
@@ -159,3 +159,60 @@ function ajaxRequest(url: string, settings: IAjaxSetttings): Promise<any> {
...
+  }
+
+  /**
+   * Resolve the underlying Promise with an optional value or another Promise.
+   */
+  resolve(value?: T | Thenable<T>): void {
+    // Note: according to the Promise spec, and the `this` context for resolve 
+    // and reject are ignored
+    this._resolve(value);
+    this._settled = true;
```
sccolbert: If we *do* maintain a settled flag, it should be set *before* resolving.

src/utils.ts
```diff
@@ -159,3 +159,60 @@ function ajaxRequest(url: string, settings: IAjaxSetttings): Promise<any> {
...
+  }
+
+  /**
+   * Reject the underlying Promise with an optional reason.
+   */
+  reject(reason?: any): void {
+    // Note: according to the Promise spec, and the `this` context for resolve 
+    // and reject are ignored
+    this._reject(reason);
+    this._settled = true;
```
sccolbert: If we *do* maintain a settled flag, it should be set *before* rejecting.

sccolbert:
Don't merge yet. I'm going to do some work against this PR with Steve.

blink1073:
@minrk, do you mind looking at the current state of the `*kernel.ts` files in this PR and let us know what you think of the API?  https://github.com/blink1073/jupyter-js-services/tree/promises_promises/src

minrk:
Yes, I'll look at it as soon as I get a chance later today.

minrk:
@blink1073 looks like the requests now only reside in helper functions. Shouldn't `Kernel.complete`, `Kernel.execute`, etc. still exist on the Kernel object as simple wrappers?

sccolbert:
@minrk This was the point I brought up about keeping the kernel interface simple. Otherwise, where do you draw the line with "convenience" methods?

src/kernelhelpers.ts
```diff
@@ -0,0 +1,200 @@
...
+import * as utils from './utils';
+
+
+/**
+ * Contents of a 'complete_request' message.
+ */
+export
+interface ICompleteRequest {
+  code: string;
+  cursor_pos: string;
```
minrk: cursor_pos is a number

minrk:
>  Otherwise, where do you draw the line with "convenience" methods?

Probably somewhere before "removing all the methods". Without them, what is the point of the Kernel object at all?

minrk:
execute isn't a "convenience method", it's the primary API of the kernel.

minrk:
Sorry, I must have misunderstood something you said the other day - I thought adding sendMessage was making the message-sending methods of the Kernel simpler, not removing the methods entirely. My understanding of the conclusion of our discussion was that `start` would be removed from Kernel to a factory function, but the rest of the methods - execute, complete, interrupt, etc. would stay.

sccolbert:
The kernel wraps the websocket connection, handles serializing the messages, and mapping the responses back to the correct callbacks. Those "methods" have nothing to do with the kernel state, so why make them methods? If you ever want alternative implementations of `IKernel`, the more methods you have on it, the harder that task becomes.

The argument is also for API consistency: why are some messages treated as special and get convenience wrappers, when others aren't? Can we have free functions for the others? Why not just make them all methods? etc...

README.md
```diff
@@ -72,3 +72,94 @@ Earlier versions may also work, but come with no guarantees.
...
+  console.log('status', status);
+});
+
+// kill the kernel
+kernel.shutdown();
+
+
+```
+
+**NoteBookSession**
```
minrk: s/B/b/

minrk:
> why are some messages treated as special and get convenience wrappers, when others aren't?

Every message should get one. This is the primary goal of the Kernel object - to present an API for sending messages.

>  If you ever want alternative implementations of IKernel, the more methods you have on it, the harder that task becomes.

Only if the implementation is different, right? Subclass and override sendMessage. This is what making the implementation of messages in sendMessage already accomplished. The other methods should be left unmodified, and makes subclasses no more complex than functions.

minrk:
Also, if you want to do it as two objects - a basic Kernel that's the inner implementation of just the wire protocol like you have here, and an outer object that presents the whole public API (i.e. just turning the functions into methods). I'm mainly expressing a preference for methods over functions for communicating with a kernel.

If the goal is to make swapping implementations easier, I think I understand some of your points better, and three objects make sense to me:

1. KernelConnection - ws messages
2. KernelResource - rest actions
3. Kernel - the actual public API, wraps KernelResource, KernelConnection

I don't care much about whether the inner implementations (the first two) are done as methods or functions - whichever seems cleaner/easier to you is fine. What I care about is the exported API, and that all the basic actions - all messages plus start/stop/interrupt - are methods on an exported class.

blink1073:
I think it is fine to have one class, I've pushed a modified version based on these comments.

minrk:
@blink1073 thanks! I prefer the existing `Kernel.foo` to `kernel.fooRequest` method naming, do you have a preference? The action is execute, which involves a request and a reply, but the public name of the action is 'execute'.

blink1073:
Fair enough, `request` is in the name of the interface object.

minrk:
(For reference: We used to have `inspect_request` etc., but replaced them with `inspect`, etc. because we decided having `_request` in the method name was a mistake.)

minrk:
> Fair enough, request is in the name of the interface object.

Makes perfect sense in the interface object, thanks!

blink1073:
Yeah, it is an overloaded term...

blink1073:
Okay, I'm going to give `Session` the same treatment now.

rgbkrk:
Just from reading what you have in the README, I'm *really excited* to work with this API.

blink1073:
@rgbkrk, the README doesn't reflect the new API yet...

rgbkrk:
Uh oh. I better read more then...

blink1073:
It will get *even better* (I hope).

blink1073:
I think the `NotebookSession` can be distilled down to this + factory functions:

```typescript
interface INotebookSession {
    /**
     * Unique id of the session.
     *
     * Read only.
     */
    id: string;

    /**
     * The path to the notebook.
     *
     * Read only.
     */
    notebookPath: string;

    /**
     * The kernel.
     *
     * Read only.
     */
    kernel: IKernel;

    /**
     * Rename the notebook.
     */
    renameNotebook(path: string): Promise<void>;
}
```

blink1073:
I had to update the `Kernel` to accommodate the `inputReply` message.  I've also finished the refactor of the `NotebookSession`.

blink1073:
This is ready for review.

minrk:
@blink1073 thanks, this is looking really good. I'll try to give it a thorough pass this afternoon.

minrk:
Awesome, merging now. I think I'll make some proposals of my own, but we can keep plugging away for now.

blink1073:
:tada:

PR #21, opened 2015-08-29
DO NOT MERGE - Discussion Only - Kernel Interface

sccolbert:
Pinging @minrk @blink1073 

Hey guys. I was looking through the kernel services code, and am still not too happy with the status quo. I think there is a large conflation of responsibilities between ajax calls and websockets, which ends up muddling state and lifetime management, and generally making the code difficult to follow.

In this PR I've stubbed out two files with just interfaces and functions: `kernel.ts` and `kernelhelpers.ts`. Ignore the diffs, and just look at those two files.

https://github.com/sccolbert/jupyter-js-services/blob/kernel-interface/src/kernel.ts
https://github.com/sccolbert/jupyter-js-services/blob/kernel-interface/src/kernelhelpers.ts

My idea is to make the kernel "object" pretty dumb and to have deal only with the websocket communication. I view the REST API as "external" kernel management, so those are expressed as external functions which either operate on kernel ids, or create IKernel objects as appropriate. I think this provides a nice mental separation of responsibilities, and it will allow for a much clear implementation of the web socket management, since it's isolated from all the other stuff.

In addition, the IKernel interface its kept extremely simple. It expose name, id, and status (read only), and a sendMessage method which takes a well-formed kernel message. We can then have a suite of functions in `kernelhelpers.ts` which assemble and all the various messages supported by the message spec. Having these as separate functions keeps the IKernel implementation small and easy to understand and test, and allows that library of helpers to expand and grow without require modifications to the kernel implementation. It also opens the door for alternative IKernel implementations, should we ever need it. Finally, it allows each function to expose an interface which makes the most sense for that particular message. For most messages, the kernel future object can be consumed internally to the function and converted into a promise result, making it much easier to use.

We could also follow a similar pattern for sessions, which would fully remove the current confusion around the various start/stop/restart/connect/disconnect methods which exist on kernel, and who should use which and when.

What do you all think about this? I much prefer this approach as I think it clearly separates concerns, and makes the code more approachable.


blink1073:
No arguments here, I find this to be much cleaner and providing room for growth in the future.  The monolithic kernel was very difficult to reason about and interact with.

rgbkrk:
As a possible external user, I primarily want a clean interface for the kernel that I can pass the websocket URL to (or a websocket for that matter), operations specific to kernels, and a way to listen for messages/events. The API calls should be separate, IMO. There can be a layer that ties these two together if you're so inclined, but I think we can have a minimal kernel library. The library for the services can also produce `Kernel`s for the user as well.

/cc @parente who I think should join these discussions, especially with respect to the alternative backends we're aiming for.

sccolbert:
@rgbkrk sounds like you are in agreement then?

minrk:
From an API consumer perspective, I think I have a Kernel, and I want to start/stop/interrupt/run code on it. I think it should be completely irrelevant to the exported API whether those actions take place on the REST API or websocket. Having the actions bundled in individual functions is fine, though.

rgbkrk:
> @rgbkrk sounds like you are in agreement then?

Gah, this means I deleted my top text! Yes, I'm in line with the idea of it.

minrk:
I'm AOK separating the inner implementation between REST and websocket. That makes sense to me. When I look at the examples, it seems like it's still that way, and quite sensible. I mainly think that splits like this probably shouldn't propagate to consumers of the public APIs of this repo (e.g. the notebook, which just wants to get and use a Kernel).

sccolbert:
Steve and I were discussing that point, and I feel that start/stop/restart a kernel are actions taken *on* a kernel by some *external* entity. After all, there is no kernel before you start one. 

On the other hand, messages sent to a kernel are intended to be processed *by* the kernel, not an external entity. So, for me, it made much more sense to separate the two APIs. Once I made that distinction everything else fell into place.

If we have external helper functions, most users will never actually call `sendMessage` (and thus not care about the implementation details). The kernel object just become an opaque handle to them.

sccolbert:
> When I look at the examples, it seems like it's still that way, and quite sensible. 

Which examples are you referring to?

> I mainly think that splits like this probably shouldn't propagate to consumers of the public APIs of this repo (e.g. the notebook, which just wants to get and use a Kernel).

So you would want a kernel object which has methods which deal with both lifecycle and messaging? Or are you saying that external users deal with a Notebook, and thus don't really care about this services repo at all?

minrk:
> Steve and I were discussing that point, and I feel that start/stop/restart a kernel are actions taken on a kernel by some external entity. After all, there is no kernel before you start one.

As long as interrupt, stop, and restart are methods on the Kernel object, that's fine. Start's the weird one to me, which makes sense to be to a function that returns a running kernel.

> So you would want a kernel object which has methods which deal with both lifecycle and messaging?

I don't view interrupt or restart as lifecycle operations, but yes, I think Kernel should have interrupt/start/stop methods. I think removing start to a function makes sense.

minrk:
I want to clarify - I think it's totally fine for these things to *be* separate, as long as the Kernel object has a simple wrapper for the REST calls other that start, since that's how I would expect ~everyone to call them.

sccolbert:
I was hoping to keep the lifecycle methods off of the kernel object, to avoid the temptation of having the kernel status attempt to reflect the lifecycle. That was a big part of the complexity of the current code. If can think of a kernel object as a thin wrapper around the websocket connection to the server-side kernel, then that status becomes more clear: it's more-or-less the status of the websocket.

minrk:
The REST vs websocket request should be considered a minor implementation detail. A Kernel object is an API for interacting with the kernel. Among the actions you take are: execute, complete, interrupt, restart. The main point for me is that if we implement something like interrupt differently ,so it is a websocket message instead of a REST request, that shouldn't affect the API at all, since the mechanism is irrelevant to the API.

minrk:
I also think it's an important distinction that interrupt and restart are not really lifecycle methods at all, in the sense of the lifecycle of the REST resource. Interrupt and restart are operations that affect the state of an existing kernel (restart does not delete and create new REST resources). The only lifecycle methods are `start` and `stop`. And it seems like an unnecessary increase in complexity for API consumers to remove `stop` from the Kernel, which seems equivalent to removing the `close` method from sockets.

sccolbert:
@minrk will you be around tomorrow for chat about this?

minrk:
Yes, I should be around tomorrow. Let's sync via gitter tomorrow morning.

sccolbert:
Sounds good. Thanks!

minrk:
Summarizing chat:

- splitting the internal implementations between websocket and REST makes sense
- leave interrupt, restart, stop on Kernel for convenience, but get rid of start.
- add utility function for sending messages on channels to reduce duplicate code (we already do this on the Python side)
- a significant amount of the state mess should be able to be cleaned up by using a factory function to create kernels whose promise only resolves when either
  1. websocket enters connected state, or 
  2. kernel_info_reply has been received

sccolbert:
thanks!

PR #22, opened 2015-09-02
use raw github for swagger url

minrk:


rgbkrk:
Definitely need this for the link to work, swagger needs the raw yaml!

blink1073:
Sweeet.

Issue #24, opened 2015-09-10
Add Support for Comm objects


blink1073:
I propose the following API for Comm Management.  It uses factory methods returning Promises, in keeping with the Kernel and Session.

http://ipython.org/ipython-doc/dev/development/messaging.html#custom-messages

```typescript
interface INewKernel extends IKernel {
  /**
   * The unhandled message signal for the kernel.
   */
  unhandledMessage: ISignal<IKernel, IKernelMessage>;

  /**
   * An unhandled comm_open message received from the client.
   */
  commOpened: ISignal<IKernel, ICommOpen>;

  /**
   * Send a shell message to the kernel.
   *
   * The future object will yield the result when available.
   */
  sendShellMessage(msg: IKernelMessage, expectReply: boolean): IKernelFuture;

  /**
   * Send a 'comm_info_request', and return the contents of the
   * 'comm_info_reply'.
   */
  commInfo(contents: ICommInfoRequest): Promise<ICommInfoReply>;

  /**
   * Connect to a comm, or create a new one.
   *
   * If a client-side comm already exists, it is returned.
   */
  connectToComm(targetName: string, commId?: string): IComm;
}

/**
 * A client side Comm interface.
 */
export
interface IComm {
  /**
   * The uuid for the comm channel.
   *
   * Read-only
   */
  commId: string;

  /** 
   * The target name for the comm channel.
   *
   * Read-only
   */
  targetName: string;

  /**
   * The onClose handler.
   */
  onClose: (data?: any) => void;

  /**
   * The onMsg handler.
   */
  onMsg: (data: any) => void;

  /**
   * Open a comm with optional data.
   */
  open(data?: any, metadata?: any): IKernelFuture;

  /**
   * Send a comm message to the kernel.
   */
  send(data: any, metadata?: any, buffers?:(ArrayBuffer | ArrayBufferView)[]): IKernelFuture;

  /**
   * Close the comm.
   */
  close(data?: any, metadata?: any): IKernelFuture;
}

/**
 * Contents of `comm_info_reply` message, a mapping of comm ids to target names.
 */
interface ICommInfoReply {
  comms: { [id: string]: string };
}

/**
 * Contents of a `comm_open` message.
 */
export
interface ICommOpen {
  comm_id: string;
  target_name: string;
  data: any;
  target_module?: string;
}
```

Note to self: use this as template for integration test: https://github.com/jupyter/notebook/blob/master/notebook/tests/services/serialize.js

[...]

PR #26, opened 2015-09-11
Add Comm Object

blink1073:
Adds the comm manager interface, using the API defined in #24.

This builds on #20.

- [x] Add unit tests
- [x] Add integration test

screen shot 2015-09-16 at 9 10 27 am


src/kernel.ts
```diff
@@ -448,6 +459,22 @@ class Kernel implements IKernel {
         future.handleMsg(msg);
       }
     }
+    if (msg.channel === 'iopub') {
+      switch(msg.header.msg_type) {
+        case 'status':
+          this._updateStatus(msg.content.execution_state);
+          break;
```
sccolbert: What about passing all `comm_` messages to a `handleCommMessage` on the `CommManager`, and letting it deal with it as it sees fit? That will remove the need for public-but-not-part-of-the-interface methods.

sccolbert:
Looks reasonable to me, but if the relationship between kernel and comm manager is 1:1, why not just merge the two? It comm is integrally tied to a kernel, might at well be part of the kernel...

blink1073:
I like the separation of concerns.  The kernel is already doing quite a bit, and handing some of that off to a domain-specific manager makes sense to me.

sccolbert:
If it's *solely* for code separation it doesn't make much sense. If *every* implementation of `IKernel` must implement their own `ICommManager`, doubly so. If the `CommManager` default implementation can be reused by any `IKernel` implementation, then that's a selling-point.

blink1073:
1) They are distinct objects on the server end.
2) I could add `comm` to `index.ts` and make `CommManager` public.

sccolbert:
so, circular dependencies are *generally* a "bad thing" (not always). That's what we have here, so we need a good reason for it. Better than "that's what it is on the server end".

sccolbert:
If the existing CommManager implementation can be reused by *any* implementation of `IKernel`, that's a good reason to separate them. But if an a custom kernel will likely need to implement its own custom comm manager, I don't see the benefit to separation. 

@minrk @jdfreder @SylvainCorlay @jasongrout Can any of you shed some light on my limited thinking?

minrk:
A CommManager implementation should be re-usable by any IKernel. It should only depend on public APIs of the kernel to send shell messages, and register callbacks for its message types. It's also reasonable to use a Kernel without a CommManager (widgets would be unavailable, but that's okay).

sccolbert:
I garner two things from that response:
1) It makes sense to keep things separated as-is.
2) We shouldn't be creating a comm manager by default.

For 2) should we make comm manager at get/set property?

minrk:
Sorry, yes. I do think it makes sense to keep the implementation separate. I don't think it's especially important that the kernel be allowed to be instantiated without having a CommManager, though.

blink1073:
If we had `registerIOPubHandler(msgType: string)` on `IKernel`, it would not have to have any notion of a `comm`.

minrk:
@blink1073 that seems like a good idea.

sccolbert:
the `iopub` channel is intended as a pub-sub pattern, you could make a signal on the kernel which is emitted when such a message is received. Like `ioPubReceived` or something of the sort. That would relieve you of having to maintain the map, and simultaneously allow for multiple consumers. The consumer would be responsible for switching on the message type.

blink1073:
`status` messages are handled by the `statusChanged` signal, but are also triggered by client side events.  IOPub messages associated with an outgoing message id are handled by the `KernelFuture`. This is meant for additional message types are that not built in to the kernel itself.

sccolbert:
remind me about this on Monday.

blink1073:
Will do.

jdfreder:
@sccolbert remind

sccolbert:
Ha! Cheers! :) 

@blink1073 and I went over it this morning. AFAIK changes are already incorporated.

blink1073:
Yep, incorporated and tested.

minrk:
@blink1073 this looks really nice. The integration test box is unchecked, do you still have more tests you want to write?

blink1073:
While setting up the integration tests, I realized I had made the false assumption that `comm*` messages were not associated with an `execute` header. The `unhandledIOPub` signal is not triggered for those messages.

blink1073:
Here is my brain dump about Comm messages:

From the client end, a comm can be created at any time
it always publishes with the current parent header from the kernel,
so it can be associated with an execution request, if desired.
https://github.com/ipython/ipykernel/blob/master/ipykernel/comm/comm.py#L79

All of the frontends to the kernel session interact with the same comm id. But, if you open a comm or send a message, the other frontends do not know about it.
https://github.com/ipython/ipykernel/blob/master/ipykernel/comm/manager.py#L108

When a frontend comes online, it will not accept messages from an existing comm connection, which is why we need the comm_info_request.

blink1073:
Based on this, I think we need something like a `commIOPub` signal.

SylvainCorlay:
You mean to notify other front-end from the creation of a comm from one of the front-ends?

sccolbert:
This is a red-flag to me (breaking into private interface): https://github.com/ipython/ipykernel/blob/master/ipykernel/comm/comm.py#L79

What's the reasoning behind including whatever the current parent header is? If it's an unsolicited comm message, just tacking on the current header doesn't make much sense to me.

blink1073:
We need to figure out a story for what a Comm *is*, it seems a bit jumbled right now.

blink1073:
It is not a 1-1 link between a client and a frontend, nor a 1-all link.

sccolbert:
Agreed. I'm still wonder why the parent header is just tacked on. That seems to be an explicit violation of the semantics of a parent header, for some convenient tracking information. It makes it more difficult to implement consumers of the messages. Maybe said tracking info can be put into metadata, so that comm messages can be handled on an entirely separate path from other messages.

blink1073:
The comm is not implemented in jupyter_console or qt_console, correct?

SylvainCorlay:
That is right -  but it could :) Even though I don't think anyone has any plan to do so.

minrk:
> What's the reasoning behind including whatever the current parent header is?

Messages coming from the kernel include the identity of the request that caused the message. A comm message from a kernel is never unsolicited, it is always the result of some execution (i.e. an execute_request).

> That seems to be an explicit violation of the semantics of a parent header, for some convenient tracking information.

I don't think so. Opening the comm is a direct consequence of the request that caused it to be opened. It's not convenience - *all* messages that come from the kernel are the result of some request, so every message has a parent.

@SylvainCorlay doesn't the widget code use comm_open's parent to identify the cell in which to place widget views?

sccolbert:
> Messages coming from the kernel include the identity of the request that caused the message. A comm message from a kernel is never unsolicited, it is always the result of some execution (i.e. an execute_request).

But if a comm message is generated from the server on a thread/timer/whatever, the parent message header could be incorrect.

minrk:
> But if a comm message is generated from the server on a thread/timer/whatever, the parent message header could be incorrect.

which is true for all messages. Nothing unique about comms, here.

sccolbert:
That sounds like it could cause some serious pain at some point.

minrk:
That's irrelevant to the protocol, though, and mainly a consequence of how IPython chooses to track the parent as global state.

sccolbert:
To me, it seems like it would be better to only include the parent header on true req-rep messages. Otherwise, "it may be a lie". Anything that's pub/sub could include its own tracking info as part of its messages.

blink1073:
What if we autodispose the Futures, and leave all IOPub as signals?

sccolbert:
If we can't trust the header for pub/sub messages, that seems the sensible approach. Let the kernel user decide how they want to handle those messages.

sccolbert:
Which means, is there really a need for the future, or can req/rep messages just return promises?

sccolbert:
(which would be really clean)

blink1073:
there are other channels besides IOPub

blink1073:
but only applicable to Execute

blink1073:
the other ones all return Promises.

blink1073:
Actually, it's just `stdin`, which could be made another signal.

blink1073:
new thought: sendShellMessage returns Promises, execute returns a Future that is connected to stdin and iopub signals and filters them.

sccolbert:
not sure I follow - could you elaborate a bit on that idea?

minrk:
@blink1073 comm_msg will have all the same semantics as execute_request in terms of IOPub messages that should be associated with the request as their parent.

blink1073:
Here's what I mean:

```typescript
interface IKernel {
    stdinReceived: ISignal<IKernel, IKernelMessage>;
    iopubReceived: ISignal<IKernel, IKernelMessage>;
    sendShellMessage(msg: IKernelMessage): Promise<IKernelMessage>;
    execute(content: IExecuteRequest): KernelFuture;
}
```

The `IKernelFuture` interface remains as-is, but instead of the `Kernel` pushing messages to it, it subscribes to the `stdinReceived` and `iopubReceived` signals, and triggers the callbacks if `msg.parent_header.msg_id` matches. 

The CommManager subscribes to `ioPubReceived`, and handles all `comm*` messages.

minrk:
@blink1073 and then leave the parent:iopub mapping up to the upstream consumer that calls execute/send? That does simplify the API at this level, certainly, though it punts the issue of connecting outputs to their requests up a level. If you did that, what would a sketch of 'request execution with handlers for its outputs' look like, since that's what ~all callers will do?

blink1073:
What I mean is, the KernelFuture would be in charge of the filters, and the consumer just hooks up the callbacks.  It would still look like this:

```typescript
var future = kernel.execute({code: 'a = 1'});
future.onIOPub = (msg) => {
   console.log('got IOPub in reply to my execute request');
}
```

sccolbert:
I think the underlying problem is that we can't reliably do the mapping. Currently, we rely on parent_header to determine routing, but the parent_header may be a lie. We need to fix that first.

sccolbert:
It also strikes me that if a comm is intended as a 1:1 channel, we should model *that*, despite how it's implemented over the message spec.

minrk:
> Currently, we rely on parent_header to determine routing, but the parent_header may be a lie.

I think trusting the parent header is the right thing to do.

> It also strikes me that if a comm is intended as a 1:1 channel, we should model that, despite how it's implemented over the message spec.

The intention of the comm channel is to provide a way to execute user code where the execution semantics are defined by user code, rather than the code-as-text of an execute request. So all the logic that associates IOPub messages with their parent should be identical for comm_msg as execute_request.

blink1073:
If I execute a cell with a Widget, and then execute another cell, then any further kernel messages to the Widget are now linked to the second cell parent_header, but still carry the correct comm_id.

minrk:
Ah, good point. You are right that comm messages should go straight to the CommManager, and not be routed *by the Kernel* based on their parent (the downstream handler will use the parent to identify the cell whose execute caused the creation of the comm). What I was referring to was that display messages should continue to be routed based on their parent, even when that parent is a comm_msg and not an execute_request.

blink1073:
Okay, we are on the same page then :smile:

minrk:
Yup, sorry. With some of this protocol stuff, it's easy to get talking exactly backwards about the relationships.

blink1073:
(ironically)

blink1073:
I still think [this option](https://github.com/jupyter/jupyter-js-services/pull/26#issuecomment-140449176) will give us the most room for growth and would relieve the Kernel from maintaining any state.

sccolbert:
Can you do a gist of what the KernelFuture implementation would look like?

blink1073:
sure, on it

minrk:
Sounds good, thanks folks.

blink1073:
https://gist.github.com/blink1073/c818f5171e9b6a24949f

blink1073:
I see now that sending a `comm*` to the Kernel should also yield a Future, right @minrk?

sccolbert:
In that gist, does the kernel directly call *any* method on the future any longer? That is, does the kernel still keep an internal map of futures?

blink1073:
No, it would not.

blink1073:
the Kernel would keep track of the last sent msgId, and use a PromiseDelegate to handle the shellPromise.

sccolbert:
ah, okay, cool - here's an updated gist which removes the kernel dependency from the future:
https://gist.github.com/sccolbert/e499eaf2f3d519c761a0

blink1073:
Sweet, thanks.  The only hole I see is that I have to expose `Comm.open` so we can get a future back.

blink1073:
So, `execute` and `comm*` will always return `Futures`, all other shell messages get a `Promise`, and the only other thing we can do is send an `inputReply`, which yields a `void`.

blink1073:
Updated the code and the API in #24, now on to update the tests.

blink1073:
This is ready for review.  I ended up reverting to the old `KernelFuture` implementation, because we can have multiple outbound shell messages before we get a reply.  Also, it alleviates the need for the `stdin` signal.

README.md
```diff
@@ -199,6 +199,58 @@ startNewSession(options).then((session) => {
...
+    var manager = new CommManager(kernel);
+      manager.connect('test').then((comm) => {
+        comm.open('initial state');
+        comm.send('test');
+        comm.close('bye');
+        done();
+      });
+    });
+  });
+});
```
sccolbert: One can return Promises from a `then` handler - which reduces indentation and makes things easier to read:

```typescript
getKernelSpecs(BASEURL).then(kernelSpecs => {
  return startNewKernel({
    baseUrl: BASEURL,
    wsUrl: WSURL,
    name: kernelSpecs.default,
  });
}).then(kernel => {
  var manager = new CommManager(kernel);
  return manager.connect('test');
}).then(comm => {
  comm.open('initial state');
  comm.send('test');
  comm.close('bye');
});
```

src/comm.ts
```diff
@@ -0,0 +1,379 @@
...
+
+  /**
+   * Open a comm with optional data.
+   */
+  open(data?: any, metadata?: any): IKernelFuture;
+
+  /**
+   * Send a comm message to the kernel.
+   */
+  send(data: any, metadata?: any, buffers?:(ArrayBuffer | ArrayBufferView)[]): IKernelFuture;
```
sccolbert: What's the purpose of returning a future here, vs relying on the `onMsg` handler to take care of any response?

src/comm.ts
```diff
@@ -0,0 +1,379 @@
...
+
+/**
+ * Contents of `comm_info` message.
+ */
+export
+interface ICommInfo {
+  /**
+   * Mapping of comm ids to target names.
+   */
+  comms: Map<string, string>;
```
sccolbert: Might be better to use an object literal here instead of a map. (more canonical) `{ [id: string]: string }`

src/comm.ts
```diff
@@ -0,0 +1,379 @@
...
+    this._kernel = kernel;
+    this._kernel.iopubReceived.connect(this._handleKernelMsg, this);
+  }
+
+  /**
+   * Connect to a comm, or create a new one.
+   *
+   * If a client-side comm already exists, it is returned.
+   */
+  connect(targetName: string, commId?: string): Promise<IComm> {
```
sccolbert: What's the reasoning behind making this return a `Promise`? It doesn't seem like the internals are actually asynchronous.

src/comm.ts
```diff
@@ -0,0 +1,379 @@
...
+   * Send a 'comm_info_request', and return the contents of the
+   * 'comm_info_reply'.
+   */
+  commInfo(targetName?: string): Promise<ICommInfo> {
+    var contents = {};
+    if (targetName !== void 0) {
+      contents = { target_name: targetName };
+    }
+    var future = sendCommMessage(this._kernel, 'comm_info_request', contents);
+    return new Promise((resolve, reject) => {
```
sccolbert: It seems like we've lost some typing here. This *should* be a compile error, because `ICommInfo.comms` is typed as a `Map`, but it's really getting an object literal. So, I think `ICommInfo` definition should be updated, but we should see where we need to add an explicit type here to get this to fail as-is.

src/comm.ts
```diff
@@ -0,0 +1,379 @@
...
+      future.onReply = (msg) => {
+        resolve(msg.content);
+      }
+    });
+  }
+
+  private _handleKernelMsg(kernel: IKernel, msg: IKernelMessage): void {
+    switch(msg.header.msg_type) {
```
sccolbert: Do these messages need to be validated?

src/comm.ts
```diff
@@ -0,0 +1,379 @@
...
+  private _onClose: (data?: any) => void = null;
+  private _onMsg: (data: any) => void = null;
+}
+
+
+/**
+ * Load a target from a module using require.js, if a module 
+ * is specified, otherwise try to load a target from the given registry.
+ */
+function loadTarget(targetName: string, moduleName: string, registry: Map<string, (comm: IComm, data: any) => any>): Promise<any> {
```
sccolbert: What should be the real return type here? I'm having trouble following what's going on.

src/comm.ts
```diff
@@ -0,0 +1,379 @@
...
+         break;
+    }
+  }
+
+  /**
+   * Handle 'comm_open' kernel message.
+   */  
+  private _handleOpen(msg: IKernelMessage): void {
+    var content = msg.content;
+    var promise = loadTarget(content.target_name, content.target_module,
```
sccolbert: This formatting is kind of hard to follow. I've massaged it a bit:

```typescript
private _handleOpen(msg: IKernelMessage): void {
  var content = msg.content;
  var promise = loadTarget(
    content.target_name, content.target_module, this._targets
  ).then((target: (comm: IComm, data: any) => any) => {
    var comm = new Comm(
      content.target_name, 
      content.comm_id, 
      this._kernel,
      () => { this._unregisterComm(content.comm_id); }
    );
    try {
      var response = target(comm, content.data);
    } catch (e) {
      comm.close();
      this._unregisterComm(comm.commId);
      console.error("Exception opening new comm");
      return Promise.reject(e);
    }
    // Regardless of the target return value, we need to
    // then return the comm
    return Promise.resolve(response).then(() => { return comm; });
  }).catch(() => {
    throw Error('Could not open comm'); 
  });
  this._comms.set(content.comm_id, promise);
}
```

But, I'm still not clear on why you have `Promise.resolve(response).then(...)` at the end, why not just return the comm?

src/comm.ts
```diff
@@ -0,0 +1,379 @@
...
+   * Handle 'comm_msg' kernel message.
+   */  
+  private _handleMsg(msg: IKernelMessage): void {
+    var content = msg.content;
+    var promise = this._comms.get(content.comm_id);
+    if (!promise) {
+      console.error('Comm promise not found for comm id ' + content.comm_id);
+      return;
+    }
+    var newPromise = promise.then((comm) => {
```
sccolbert: We're going to need another approach to storing the mapping of comms. Running through a promise to dispatch each message will be way too expensive, but in CPU cycles and GC churn.

src/kernelfuture.ts
```diff
@@ -0,0 +1,212 @@
+// Copyright (c) Jupyter Development Team.
```
sccolbert: Why did you move this out of `kernel.ts` instead of just exporting the future implementation?

src/comm.ts
blink1073: Comm messages are expected to cause side effects, and are treated like an `execute` call.

src/comm.ts
blink1073: A `Comm` created on the server side is asynchronous, because it my have to make an AMD load, so we treat everything on `CommManager` as asynchronous.

src/comm.ts
sccolbert: How is that different from onMsg?

src/comm.ts
blink1073: I think this should be a `Promise<void>`, I'll change it.

blink1073: Nevermind, it was supposed to be a callback: `Promise<(comm: IComm, data: any) => any>`

src/comm.ts
blink1073: `onMsg` is for an explicit incoming  `comm_msg`, but any other type of IOPub data can be triggered.

blink1073:
I updated the way kernel promises are handled.  We only store it internally as a promise if we get a `comm_open` message until it gets resolved.  So, the `comm_msg` Promise thrashing will occur until the comm is ready.

src/comm.ts
sccolbert: what's the expected reply when sending a comm message?

I'm harping on this because I imagine comm messages to be sent *much* more rapidly than normal messages. Every time you move a slider widget, a comm message will be sent. So, I'd like that path to be as efficient as possible.

src/comm.ts
```diff
@@ -0,0 +1,404 @@
...
+    if (promise) {
+      return promise;
+    }
+    var comm = this._comms.get(commId);
+    if (!comm) {
+      comm = new Comm(targetName, commId, this._kernel, () => {
+        this._unregisterComm(comm.commId);
+      });
+    }
+    this._comms.set(commId, comm);
```
sccolbert: this should be inside the `if` clause.

src/comm.ts
blink1073: There is no expected reply to a comm message.

src/validate.ts
```diff
@@ -2,12 +2,70 @@
...
 
 import { INotebookId, ISessionId } from './isession';
 
 
 /**
+ * Validate an Kernel Message as being a valid Comm Message.
+ */
+export 
+function validateCommMessage(msg: IKernelMessage): boolean {
+  var fields = ['comm_id', 'target_name'];
```
sccolbert: allocate this array statically, instead of per-invocation

src/validate.ts
```diff
@@ -2,12 +2,70 @@
...
+    if (typeof msg.content[fields[i]] !== 'string') {
+      return false;
+    }
+  } 
+  return true;
+}
+
+
+function validateKernelHeader(header: any): void {
+  var fields = ['username', 'version', 'session', 'msg_id',  'msg_type'];
```
sccolbert: allocate this array statically, instead of per-invocation

src/validate.ts
```diff
@@ -2,12 +2,70 @@
...
+  }
+}
+
+
+/**
+ * Validate an object as being of IKernelMessage type.
+ */
+export
+function validateKernelMessage(msg: IKernelMessage) : void {
+  var fields = ['header', 'parent_header', 'metadata', 'content',
```
sccolbert: allocate this array statically, instead of per-invocation

src/comm.ts
blink1073: https://ipython.org/ipython-doc/dev/development/messaging.html#custom-messages

"These messages are fully symmetrical - both the Kernel and the Frontend can send each message, and no messages expect a reply"

src/comm.ts
```diff
@@ -0,0 +1,404 @@
...
+    return Promise.resolve(comm);
+  }
+
+  /**
+   * Register the handler for a "comm_open" message on a given targetName.
+   */
+  registerTarget(targetName: string, cb: (comm: IComm, data: any) => void): void {
```
sccolbert: This might be better names `setTargetHandler` or something other than `register`. To me, "register" denotes "possibly more than one", whereas this is actually a 1:1 mapping.

src/comm.ts
```diff
@@ -0,0 +1,404 @@
...
+       case 'comm_close':
+         this._handleClose(msg);
+         break;
+    }
+  }
+
+  /**
+   * Handle 'comm_open' kernel message.
+   */  
+private _handleOpen(msg: IKernelMessage): void {
```
sccolbert: Indentation for this method is incorrect (there is none).

src/comm.ts
```diff
@@ -0,0 +1,404 @@
...
+  this._commPromises.set(content.comm_id, promise);
+}
+
+  /**
+   * Handle 'comm_close' kernel message.
+   */  
+  private _handleClose(msg: IKernelMessage): void {
+    var content = msg.content;
+    var promise = this._commPromises.get(content.comm_id);
+    if (!promise) {
```
sccolbert: you can make yourself a helper method to lookup/get the comm promise.

src/comm.ts
```diff
@@ -0,0 +1,404 @@
...
+      }
+      promise = Promise.resolve(comm);
+    }
+    promise.then((comm) => {
+      this._unregisterComm(comm.commId);
+      try {
+        comm.close(msg.content.data);
+      } catch (e) {
+        console.log("Exception closing comm: ", e, e.stack, msg);
+      }
```
sccolbert: although, is there a reason this needs to be handled as a promise, other than uniformity? (which is an okay answer)

src/comm.ts
```diff
@@ -0,0 +1,404 @@
...
+      var newPromise = promise.then((comm) => {
+        try {
+          var onMsg = comm.onMsg;
+          if (onMsg) onMsg(msg.content.data);
+        } catch (e) {
+          console.log("Exception handling comm msg: ", e, e.stack, msg);
+        }
+        return comm;
+      });
+      this._commPromises.set(content.comm_id, newPromise);
```
sccolbert: there's no need to reset the promise here. You can `.then` a promise multiple times.

src/comm.ts
```diff
@@ -0,0 +1,404 @@
...
+  /**
+   * Construct a new comm channel.
+   */
+  constructor(targetName: string, commId: string, kernel: IKernel, callback: () => void) {
+    super(callback);
+    this._target = targetName;
+    this._id = commId;  
+    this._kernel = kernel;
+  }
+
```
sccolbert: clear internal state on dispose?

src/comm.ts
blink1073: Promise :all the things:!

src/comm.ts
sccolbert: Hmm, that's no bueno for this situations. Particularly because `autodispose` is being set to `false` on the future. The most likely thing to happen, is that the consumer ignores the return value of sending on a comm, and inadvertently leaks an insane amount of futures.

blink1073: Looks like we may need to add an LRU cache after all....

blink1073:
Ready for review @minrk.

minrk:
Awesome, thanks for the ping.

src/kernel.ts
```diff
@@ -540,6 +780,13 @@ function restartKernel(kernel: IKernel, baseUrl: string): Promise<void> {
 }
 
 
+interface ICommPayload {
+  msgType: string;
+  contents: any;
```
minrk: we generally use singular 'content'

src/ikernel.ts
```diff
@@ -251,7 +314,7 @@ interface IKernel {
    *
    * The future object will yield the result when available.
    */
-  sendShellMessage(msg: IKernelMessage): IKernelFuture;
+  sendShellMessage(msg: IKernelMessage, expectReply: boolean): IKernelFuture;
```
minrk: Should `expectReply` be optional, with a default of `False`?

blink1073: Yeah, makes sense.

blink1073:
@minrk, comments addressed.

minrk:
Thanks, merging and marching on!

PR #27, opened 2015-09-14
make utils.uuid a random lowercase string

minrk:
We don't have any need for formal UUIDs, just sufficiently random strings.

This was mainly prompted by the caps letters looking shouty.


blink1073:
IF YOU INSIST

PR #29, opened 2015-09-25
Add Missing Tests

blink1073:
Relies on https://github.com/jupyter/notebook/pull/493 to allow all request types for allowed an cross-origin address.

- Use dev version of notebook
- Add tests for `Kernel.shutdown`, `Session.shutdown`, and `Session.renameNotebook`

### DIFF
--- a/scripts/travis_install.sh
+++ b/scripts/travis_install.sh
@@ -8,9 +8,9 @@ conda config --set always_yes yes --set changeps1 no
 conda update -q conda
 conda info -a
 
-# install development version of juptyer
+# install development version of notebook
 # https://github.com/jupyter/notebook/blob/master/.travis.yml
 git clone https://github.com/jupyter/notebook notebook-dev
 cd notebook-dev
 git clone --quiet --depth 1 https://github.com/minrk/travis-wheels travis-wheels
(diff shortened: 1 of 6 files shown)


blink1073:
ping @minrk

scripts/travis_install.sh
```diff
@@ -7,4 +7,10 @@ hash -r
...
 conda update -q conda
 conda info -a
-conda install jupyter
+
+# install development version of juptyer
+# https://github.com/jupyter/notebook/blob/master/.travis.yml
+git clone https://github.com/jupyter/notebook notebook-dev
+cd notebook-dev
+git clone --quiet --depth 1 https://github.com/minrk/travis-wheels travis-wheels
+pip install -f travis-wheels/wheelhouse --pre file://$PWD#egg=notebook[test] coveralls
```
minrk: `--pre` shouldn't be needed here. Further, since we aren't running the notebook tests, this should be able to be:

    pip install -f travis-wheels/wheelhouse .

blink1073:
Fixed.

minrk:
@blink1073 do you want to squash those debugging commits, or keep them in there?

blink1073:
Squashed.

minrk:
Thanks!

PR #35, opened 2015-10-05
Allow options to ajax request calls

### LINKED ISSUE #31: Allow xhr options to be passed (e.g., withCredentials)
Currently, XHR calls such as the one here (https://github.com/jupyter/jupyter-js-services/blob/master/src/kernel.ts#L105) and in jupyter/notebook do not allow the caller to specify additional XHR options. If I want to run my kernel provider (e.g., tmpnb) behind a proxy that, for example, requires auth headers to be sent along with XHR requests, I have to hack the code to set the withCredentials flag.  Likewise, if I have a setup in which I need to send additional headers, I have to hack the code. It would be nice if these comm options could be set via args to the functions.

Related: oreillymedia/thebe#52

blink1073:
Fixes #31, adding optional Ajax parameters to all methods and functions with Ajax calls.

```typescript
/**
 * Options for AJAX calls.
 */
export
interface IAjaxOptions {
  timeout?: number;
  requestHeaders?: { [key: string]: string; };
  async?: boolean;
  withCredentials?: boolean;
  user?: string;
  password?: string;
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "postinstall": "npm dedupe",
     "prepublish": "npm run build",
     "test": "mocha test/build/test*.js",
+    "test:debug": "pkill node; node-debug _mocha test/build/test*.js",
     "test:integration": "cd test && python run_test.py",
     "test:coverage": "istanbul cover _mocha -- test/build/test*.js"
   },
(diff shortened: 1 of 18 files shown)


blink1073:
Ready for review.

blink1073:
cc @parente.

minrk:
That's a lot of `ajaxOptions`!

PR #39, opened 2015-10-08
[WIP] Travis Fixes

blink1073:
Fixes intermittent errors on Travis that were a result of race conditions.
We now trigger a `kernelInfo` on restart, to ensure we get a response back.


blink1073:
Fixed by #42.

PR #45, opened 2015-10-16
allow callers to take control of KernelFuture disposal

minrk:
enables cases where callers may want to handle out-of-order output, and have a bounded number of futures to keep around (e.g. 1 cell = 1 future in the notebook).

A simpler option might be to make this non-optional on execute and comm_msg, and just require callers to be in charge of disposal in those cases.


blink1073:
LGTM.

PR #47, opened 2015-10-20
Fix and document Kernel lifecycle handling

### LINKED ISSUE #40: Kernel Scope and Lifetime API

blink1073:
Fixes #40.

### DIFF
--- a/src/kernel.ts
+++ b/src/kernel.ts
@@ -35,6 +35,12 @@ var KERNEL_SERVICE_URL = 'api/kernels';
 var KERNELSPEC_SERVICE_URL = 'api/kernelspecs';
 
 
+/**
+ * The error message to send when the kernel is not ready.
+ */
+var KERNEL_NOT_READY_MSG = 'Kernel is not ready to send a message';
+
+
 /**  
  * Fetch the kernel specs.
  *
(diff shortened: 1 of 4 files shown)


blink1073:
ping @jasongrout, based on our discussions last week.

src/kernel.ts
```diff
@@ -422,6 +459,9 @@ class Kernel implements IKernel {
    * received and validated.
    */
   kernelInfo(): Promise<IKernelInfo> {
+    if (this._status === KernelStatus.Dead) {
+      return Promise.reject(Error('Kernel is not ready to send a message'));
+    }
```
minrk: Shouldn't these status checks be in sendKernelMessage (like you already have done in sendShellMessage) to avoid the duplication?

blink1073: This one could be, but the others have a more stringent definition of "ready to send".

minrk: Ah, I see. Unfortunate, but not a big deal. +1, then.

blink1073:
:beers:

PR #54, opened 2015-11-04
provide default values for baseUrl, wsUrl

minrk:
In existing use cases, wsUrl can never deviate from the default value, so it doesn't make sense to require it to be specified.


blink1073:
Min, I'm traveling, the test failure is likely due to Miniconda not including pip, you need to update travis_install as in muly other open PR. 



—

blink1073:
Great, thanks Min.

PR #62, opened 2015-12-18
API Upgrades

### LINKED ISSUE #57: ajaxRequest should warn on invalid/undefined `settings.methods`
And generally I would have expected the API to be closer to $.ajax the setting/option distinction is bit annoying to refactor when we already have layers of call in JS. 

Here I was caught by GET not being the default.

### LINKED ISSUE #60: Add a cache option
https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest#Bypassing_the_cache

blink1073:
- Improves the API of `ajaxRequest` (Fixes #57).
- Disables cache on the same methods as the current notebook (Fixes #60).
- Moves `ajaxSettings` to the object level and makes them optional, to account for implementations that may not use ajax, or consumers of the object that would not have access to the proper settings otherwise.
- Cleans up the API for functions in `kernel.ts` and `session.ts`, which now take `IKernelOptions` and `ISesssionOptions` across the board.
- Renames `IContent` to `IContentsManager` for clarity.
- Adds `IKernelManager` and `INotebookSessionManager` and associated implementations that abstract the `baseUrl` and `ajaxSettings` for a consumer.

cc @parente @Carreau @minrk


blink1073:
This is ready for review.  I will be dog-fooding it tomorrow in a `jupyter-js-filebrowser` branch.

blink1073:
I set this back to WIP because the new `*Manager` classes need tests.

blink1073:
Finished and dog-fooded in https://github.com/jupyter/jupyter-js-filebrowser/pull/25.

blink1073:
I've been using this without issue for the past three weeks, any objections to merging?

parente:
Doesn't look like anything will impact us over in dashboards. LGTM.

jasongrout:
I just realized the npm version already had this merged.

blink1073:
Yes, I needed it to continue work on the filebrowser.

jasongrout:
Thanks!  Given @parente's +1, and I glanced at it too, let's merge.

PR #66, opened 2016-01-13
Update comm and kernel future to use full message

### LINKED ISSUE #64: Use the full message in the Comm methods instead of just the data

blink1073:
Fixes #64.


minrk:
Nice!

PR #74, opened 2016-01-20
Execute defaults

jasongrout:
This brings the execute command defaults into alignment with the message spec documentation at http://jupyter-client.readthedocs.org/en/latest/messaging.html#execute.

@ellisonbg - you changed the defaults in the kernel.js code away from these in https://github.com/jupyter/notebook/commit/f306423c7ea3bc60735a8bfb4f5f2cc60dc36b97.  Do you recall why you did this, and do you have comments about changing it back in jupyter-js-services?

### DIFF
--- a/src/kernel.ts
+++ b/src/kernel.ts
@@ -633,10 +633,11 @@ class Kernel implements IKernel {
       session: this._clientId
     }
     var defaults = {
-      silent : true,
-      store_history : false,
+      silent : false,
+      store_history : true,
       user_expressions : {},
-      allow_stdin : false
+      allow_stdin : true,
+      stop_on_error : false
     };
     contents = utils.extend(defaults, contents);
     var msg = createKernelMessage(options, contents);


jasongrout:
I'll just add that the current kernel.js execute function has contradicting documentation about the default of silent (it says it is false in one place and true in another): https://github.com/jupyter/notebook/blob/7e4f72ed6c4a8c0857d16cee648f9e692debf33c/notebook/static/services/kernels/kernel.js#L700

src/kernel.ts
```diff
@@ -633,10 +633,11 @@ class Kernel implements IKernel {
       session: this._clientId
     }
     var defaults = {
-      silent : true,
-      store_history : false,
+      silent : false,
+      store_history : true,
       user_expressions : {},
-      allow_stdin : false
+      allow_stdin : false,
```
blink1073: This is listed as `true` here: http://jupyter-client.readthedocs.org/en/latest/messaging.html#execute

jasongrout: Yes, thanks, good catch!

blink1073:
I'm in favor of merging this, and reverting if Brian has evidence why we should.

PR #78, opened 2016-01-22
Add a comm target registry

blink1073:
cc @jdfreder @jasongrout @SylvainCorlay

### DIFF
--- a/src/ikernel.ts
+++ b/src/ikernel.ts
@@ -560,11 +560,6 @@ interface IKernel extends IDisposable {
   unhandledMessage: ISignal<IKernel, IKernelMessage>;
 
   /**
-   * A signal emitted for unhandled comm open message.
-   */
-  commOpened: ISignal<IKernel, IKernelMessage>;
-
-  /**
    * The id of the server-side kernel.
    *
    * #### Notes
@@ -770,9 +765,9 @@ interface IKernel extends IDisposable {
    * #### Notes
    * Only one comm target can be registered at a time, an existing
    * callback will be overidden.  A registered comm target handler will take
-   * precidence over a comm which specifies a `target_module`.
+   * precedence over a comm which specifies a `target_module`.
    */
-  registerCommHandler(targetName: string, callback: (msg: IKernelMessage) => void): IDisposable;
+  registerCommTarget(targetName: string, callback: (comm: IComm, msg: IKernelMessage) => void): IDisposable;
 
   /**
    * Optional default settings for ajax requests, if applicable.
--- a/src/ikernel.ts
+++ b/src/ikernel.ts
@@ -759,6 +759,21 @@ interface IKernel extends IDisposable {
   connectToComm(targetName: string, commId?: string): IComm;
 
   /**
+   * Register a comm target.
+   *
+   * @param targetName - The name of the comm target.
+   *
+   * @param callback - The callback invoked for a comm open message.
+   *
+   * @returns A disposable used to unregister the com target.
+   *
+   * #### Notes
+   * Only one comm target can be registered at a time, an existing
+   * callback will be overidden.
+   */
+  registerCommTarget(targetName: string, callback: (msg: IKernelMessage) => void): IDisposable;
+
+  /**
    * Optional default settings for ajax requests, if applicable.
    */
   ajaxSettings?: IAjaxSettings;
--- a/src/ikernel.ts
+++ b/src/ikernel.ts
@@ -268,6 +268,26 @@ function isClearOutputMessage(msg: IKernelMessage): msg is IKernelIOPubClearOutp
 
 
 /**
+ * IOPub comm_open kernel message specification.
+ *
+ * See [Comm open](http://jupyter-client.readthedocs.org/en/latest/messaging.html#opening-a-comm).
+ */
+export
+interface IKernelIOPubCommOpenMessage extends IKernelMessage {
+  content: ICommOpen;
+}
+
+
+/**
+ * Check if an IKernelMessage is an iopub comm_open message.
+ */
+export
+function isCommOpenMessage(msg: IKernelMessage): msg is IKernelIOPubCommOpenMessage {
+  return msg.header.msg_type === "comm_open";
+}
+
+
+/**
  * Kernel information specification.
  *
  * See [Messaging in Jupyter](http://jupyter-client.readthedocs.org/en/latest/messaging.html#kernel-info).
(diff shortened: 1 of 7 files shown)


src/kernel.ts
```diff
@@ -884,12 +905,15 @@ class Kernel implements IKernel {
...
     }
     var content = msg.content as ICommOpen;
-    if (!content.target_module) {
-      this.commOpened.emit(msg);
+    this.commOpened.emit(msg);
+    var targetName = content.target_name;
+    var moduleName = content.target_module;
+    if (this._targetRegistry[targetName]) {
+      this._targetRegistry[targetName](msg);
+    }
```
blink1073: How do we want to handle the case there there is a `moduleName` *and* the `targetName` is registered?

jdfreder: The widgets has a few registries scattered about, and all of the registries take precedence over the differed loading mechanism.  Not to say we have to do that here, but my reasoning was it made it easy for the user to come along and override something internal, during live exploration in the notebook.

blink1073: So, your vote is for registry trumps `moduleName`?

jdfreder: Yup

blink1073: That would be my preference as well, let's go with it.

blink1073: API updated, I'll add tests.

jasongrout: I'm not so sure.  You could consider (modulename, targetname) more specific than just targetname.  On the other hand, actively registering a function feels more specific than looking up something that is installed.  I think we need to standardize on one way for our stuff, and leave the other option for easy customization.

blink1073: The modulename is provided by the kernel, which has no way of knowing the status of the front end.  I'd say that manually registering on the front end means one has *more* knowledge, and therefore takes precedence.

jasongrout: Another way to look at it: Making registry trump the loader means that the targetname namespace is essentially a global namespace.  Making loader trump registry means that empty modulename namespace is just *one* of the possible module namespaces.

If we have two modules that provide the same targetnames, how do we override both at the same time?  We can't if the targetname trumps loader. In fact, overriding one automatically overrides the other.  That seems problematic, if we think of the registry as an override for module namespaces.

blink1073: Do we ever want two front end Comms talking to the same comm_id?

jasongrout: I think not.  The whole idea is a private communication channel, so I think we can insist on a multiplexer if you want multiple people talking on the same comm channel.

jasongrout: @blink1073 - that's right about registering meaning you have more knowledge.  Maybe a registration should have a `(modulename, targetname)` pair?

jasongrout: Sylvain and I are talking, and he's convincing me that maybe we don't want *any* registry, and just have `(module, target)` based loaders...

blink1073: That would require the kernel to explicitly know the target + module, right?

jasongrout: Okay, so Sylvain and I think it would be better to get rid of the registry, *always* use the static `(module, target)` based loaders, and eventually provide an easy way to hack into the general import/loader semantics to dynamically load code

jdfreder: I see real a benefit using to a registry based approach, at least in ipywidgets, because it allows the user to hack on things in the live notebook.  It's also guaranteed to work as long as we support it, where methods to define arbitrary modules, like `define('mymodule', function() {})` provided by require.js, are not guaranteed to exist and depend on the loader we use in the frontend.

jasongrout: (For future readers) We discussed on appear.in, and decided to:

1. mimic the existing functionality in the existing load_class: https://github.com/jupyter/notebook/blob/a8f6265d07378d7e96d63cb77c2641de52cd3a4a/notebook/static/base/js/utils.js#L701
2. Propose an messaging spec change to make the comm target_module mandatory, deprecate the existing registry.
3. Provide a way to override system import paths to override things/load things dynamically.

blink1073:
I released a new `jupyter-js-utils` with `loadClass`, and updated the API here.

jdfreder:
Sweet

src/kernel.ts
```diff
@@ -1023,6 +1032,7 @@ class Kernel implements IKernel {
   private _commPromises: Map<string, Promise<IComm>> = null;
   private _comms: Map<string, IComm> = null;
   private _isWebSocketClosing = false;
+  private _targetRegistry: { [key: string]: (comm: IComm, msg: IKernelMessage) => void; } = Object.create(null);
```
jasongrout: Cool, I didn't know about `Object.create(null)`.  Any reason we're not using a Map here?

src/ikernel.ts
```diff
@@ -759,6 +754,22 @@ interface IKernel extends IDisposable {
...
+   * @param callback - The callback invoked for a comm open message.
+   *
+   * @returns A disposable used to unregister the comm target.
+   *
+   * #### Notes
+   * Only one comm target can be registered at a time, an existing
+   * callback will be overidden.  A registered comm target handler will take
+   * precedence over a comm which specifies a `target_module`.
+   */
+  registerCommTarget(targetName: string, callback: (comm: IComm, msg: IKernelMessage) => void): IDisposable;
```
jasongrout: Can we subclass the IKernelMessage to make an ICommOpenMessage that is passed to the callback, to give some extra type checking information?  Something like this in the ikernel.ts file:

```

/**
 * IOPub comm_open kernel message specification.
 * 
 * See [Comm open](http://jupyter-client.readthedocs.org/en/latest/messaging.html#opening-a-comm).
 */
export
interface IKernelIOPubCommOpenMessage extends IKernelMessage {
  content: {
    comm_id: string;
    target_name: string;
    data: any;
    target_module?: string;
  }
}

/**
 * Check if an IKernelMessage is an iopub comm_open message.
 */
export
function isCommOpenMessage(msg: IKernelMessage): msg is IKernelIOPubCommOpenMessage {
  return msg.header.msg_type === "comm_open";
}
```

src/kernel.ts
```diff
@@ -911,12 +921,11 @@ class Kernel implements IKernel {
           comm.close();
           this._unregisterComm(comm.commId);
```
jasongrout: should we just call `comm.dispose()` here?

src/kernel.ts
blink1073: I find this type of map easier to work with in practice.

blink1073: But, everything else there is already a Map, so I'll be consistent.

blink1073: Ah, but `loadClass` is expecting this type of map, so I'll keep it as a regular object.

src/kernel.ts
blink1073: Okay, just `close()` it is.

src/ikernel.ts
blink1073: Done.

jasongrout:
These changes look great to me.  I think we should merge and iterate, if needed.

jasongrout:
@blink1073 - since this is a breaking change, perhaps bumping to 0.5.0 is in order?

PR #79, opened 2016-01-26
iopub errors don't have execution_count

jasongrout:
Turns out that I read the spec wrong when writing the iopub message validation code, and iopub error messages don't have execution_count fields.


blink1073:
All righty.

PR #81, opened 2016-01-26
Add shutdown reply iopub message

blink1073:


minrk:
From https://github.com/jupyter/jupyter_client/pull/113, I think this shouldn't be added. The client should log this as an unhandled message and move on. Is this not what happens?

blink1073:
Right now it is an error, I'll open a PR to make it just log.

blink1073:
Actually, not needed, I was only seeing a problem because I had "Pause on caught exceptions", the current behavior is correct.

minrk:
Great, that's what the code looked like to me.

PR #101, opened 2016-03-09
Add the ability to change the kernel in a session

blink1073:
Augments the `INotebookSession` to allow the kernel to change.
Also refactored the tests so they are not all importing from `test/src/kernel`.  This allows one to run an individual test file separately.


test/src/utils.ts
```diff
@@ -7,20 +7,34 @@ import expect = require('expect.js');
...
+import {
+  MockXMLHttpRequest
+} from './mockxhr';
 
 
 // stub for node global
 declare var global: any;
 
 
+overrideWebSocket();
```
dwillmer: Not 100% sure what this is for.

blink1073: It sets up the global WebSocket to be our mock object.

PR #110, opened 2016-03-14
Use PATCH to change session kernel

blink1073:
Relies on https://github.com/jupyter/notebook/pull/1206.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ Earlier versions may also work, but come with no guarantees.
 
 Starting the Notebook Server
 ----------------------------
-The library requires a running Jupyter Notebook server, v4.1+, launched as:
+The library requires a running Jupyter Notebook server (development version), launched as:
 
 ```bash
 python -m notebook --NotebookApp.allow_origin="*"
--- a/README.md
+++ b/README.md
@@ -362,6 +362,7 @@ contents.listCheckpoints("/foo/bar.txt").then((models) => {
 **Configuration**
 
 ```typescript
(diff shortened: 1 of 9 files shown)


minrk:
Merged upstream, if you want to bump things here. It's a little disconcerting that the tests are passing prior to the upstream patch being merged. Is this behavior not tested?

blink1073:
I didn't have a way to test agaist the notebook server itself, only the logic.  I'll add that test.

blink1073:
It would be cleaner to have a `changeKernel(name: string)` and a `connectToKernel(id: string)` instead, what do you think, @minrk?

minrk:
where connectToKernel produces a new session, while changeKernel always gets a new kernel? That leaves out `changeKernel(kernel_id)`, where you have a notebook running and want to attach it to a running kernel.

blink1073:
No, what I *really* want to do is add `kernelId` to `ISessionOptions` and then have one function called `NotebookSession.changeKernel(options: ISessionOptions)`, and use the `kernelId` when starting a new session.  So, maybe I'll make the PR against notebook first to do that.

blink1073:
This part is done.  I'll add the ability to connect to an existing session after https://github.com/jupyter/notebook/pull/1237 is merged.

minrk:
Merged that, but this needs a rebase now. Will merge when it applies.

blink1073:
Ready to roll.

minrk:
rolling :bowling:

PR #115, opened 2016-03-18
Return kernel before websocket connection

### LINKED ISSUE #114: Return kernel before websocket connection
This makes sense now that messages can be queued.

blink1073:
Fixes #114.


afshin:
:+1:

PR #125, opened 2016-04-06
Add contentType handling for kernel ajax methods

### LINKED ISSUE #123: Set the contentType for ajaxRequest
Example: The POST request to /api/kernels currently has `Content-Type: text/plain`. This is not a problem for the notebook server which does not care, but becomes problematic for alternative backends (dashboard server) that want deal with requests according to content type (express body parser).

Probably should be done throughout. I believe all of the RESTful endpoints take JSON.

blink1073:
Fixes #123.


afshin:
👍

PR #129, opened 2016-04-11
Move kernelfuture to a separate file

blink1073:
This allows us to use it for mock implementations of the kernel and the session.


afshin:
👍

PR #136, opened 2016-04-27
Update IHistoryRequest interface to reflect API.

afshin:
Just a minor update to the `IHistoryRequest` interface to reflect the API requirements:

http://jupyter-client.readthedocs.org/en/latest/messaging.html#history


src/ikernel.ts
```diff
@@ -15,6 +15,10 @@ import {
 } from 'jupyter-js-utils';
 
 
+export
+type hist_access_type = 'range' | 'tail' | 'search';
```
blink1073: You'll need to update the required Typescript and use this version of TypeDoc: "https://github.com/SierraSoftworks/typedoc.git#cb5aea2b218b4f773451e2818a48629ee9c5066e"

blink1073: Also, why not use camelCase here?

blink1073: The underscore case was to match the spec, but the variable name doesn't need to.

blink1073:
:+1:

PR #140, opened 2016-05-05
Make each instance of a kernel and a session unique and add iopub signal

blink1073:
This allows multiple client connections to connect to a given session or kernel with a unique client session id, since the server side websockets are one-to-one for client session ids.

Also adds an iopub signal, and only using the unhandledmessage signal for messages that are not iopub and match the current client session id.

cf https://github.com/jupyter/jupyter-js-notebook/issues/166#issuecomment-215853776

### DIFF
--- a/src/ikernel.ts
+++ b/src/ikernel.ts
@@ -613,6 +613,11 @@ interface IKernel extends IDisposable {
   statusChanged: ISignal<IKernel, KernelStatus>;
 
   /**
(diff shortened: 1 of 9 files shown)


blink1073:
I'm going to follow this up with another PR adding an iopub signal, and only using the unhandledmessage signal for messages that are not iopub and match the current client session id.

afshin:
👍

PR #141, opened 2016-05-06
Use updates from server

blink1073:
When fetching session information from the server, update existing client side instances.
Also adds a `notebookPathChanged` signal on the session.
The intent is that one part of the application will call `listSessions` periodically, which will update the information of all running sessions.

cf https://github.com/jupyter/jupyter-js-notebook/issues/166#issuecomment-215853776


afshin:
Looks good, 👍

PR #150, opened 2016-06-23
Clean up interfaces

blink1073:
Puts the kernel interfaces in `IKernel` and `KernelMessage` namespaces, and puts the session interfaces in an `ISession` interface and marks everything as a JSON object.

Renames `NotebookSession -> Session`.

Temporarily add `json.ts` here until we switch to the new `phosphor` library.

Readmes and examples updated and verified.


blink1073:
Probably best viewed as the new file content instead of a diff.

src/ikernel.ts
```diff
@@ -14,1042 +14,1020 @@ import {
...
+} from './json';
 
 
 /**
- * The options object used to initialize a kernel.
+ * A namespace for kernel types, interfaces, and type checker functions.
  */
-export
-interface IKernelOptions {
+export namespace kernel {
```
afshin: Just a line break:
```
export
namespace kernel
```

afshin:
Good use of the new JSON definitions, it makes these interfaces more accurate and type safe.

blink1073:
Word.

blink1073:
This is ready for review.  The substantial changes are in `ikernel.ts` and `isession.ts`, with the rest of the changes being the necessary modifications.

afshin:
This is epic. It looks good, cleaner organization, better interfaces, and code conventions.

👍

PR #152, opened 2016-06-26
Fix handling of execute reply and add execute request options

blink1073:


src/ikernel.ts
```diff
@@ -1047,34 +1047,112 @@ namespace KernelMessage {
...
+
+    /**
+     * Whether to execute the code as quietly as possible.
+     * The default is `false`.
+     */
     silent?: boolean;
+
+    /**
+     * Whether to store history of the execution.
+     * The default `true` is silent is False.
```
minrk: s/is silent/if silent/

src/ikernel.ts
```diff
@@ -1047,34 +1047,112 @@ namespace KernelMessage {
...
+     * It is forced to  `false `if silent is `true`.
+     */
     store_history?: boolean;
+
+    /**
+     * A mapping of names to expressions to be evaluated in the
+     * user's dict.
```
minrk: s/user's dict/kernel's interactive namespace/

src/ikernel.ts
blink1073: Fixed

src/ikernel.ts
blink1073: Fixed

minrk:
Thanks!

PR #161, opened 2016-06-30
Add a published webpack bundle

### LINKED ISSUE #160: Make a single .js file that can be included in other projects
We would like to use the jupyter-js-services library, however it would be very complicated to include an ```npm build``` style workflow into our project. 

It would be great if there was a single ```jupyter-js-services.js``` that file that we could just include into our html page without having to include additional libraries. Something like [webpack](https://webpack.github.io/) or [browserify](http://browserify.org/) can do this. If we do this on our end, it means a significant effort every time we want to update to the newest version, so we were hoping that you could generate single such a file somewhere in this repository. 

/cc @florian-rabe @minrk

blink1073:
Fixes #160.

cc @tkw1536 @SylvainCorlay

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,8 +16,9 @@
     "requirejs": "^2.2.0",
     "rimraf": "^2.4.2",
     "text-encoding": "^0.5.2",
-    "typedoc": "https://github.com/SierraSoftworks/typedoc.git#cb5aea2b218b4f773451e2818a48629ee9c5066e",
+    "typedoc": "^0.4.4",
     "typescript": "^1.8.0",
+    "webpack": "^1.13.1",
     "ws": "^1.0.1",
     "xmlhttprequest": "^1.8.0"
   },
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
LGTM!

SylvainCorlay:
@tkw1536, once a version is published with this patch, you should be able to amd-require (requirejs) 

`https://npmcdn.com/jupyter-js-services/dist/index.js`

or even add a semver requirement, e.g.

`https://npmcdn.com/jupyter-js-services@^0.12.0/dist/index.js`

blink1073:
Published as `jupyter-js-services@0.11.4`.

SylvainCorlay:
OK, fixed the URL above.

SylvainCorlay:
Hum, actually, I think that npmcdn prefers the main js file to be `index.js`

SylvainCorlay:
so that you can only require
`https://npmcdn.com/jupyter-js-services@0.11.4/dist/`

SylvainCorlay:
I don't really know if it is important. We could probably decide for a pattern at some point.

It would be really cool that one could make a pure cdn-hosted client to jupyter with npmcdn.

blink1073:
Yeah, whatever we do it should be consistent.

minrk:
Awesome, thanks @blink1073! We should probably do the same for jupyterlab at some point.

blink1073:
Yep, that is the next target ;).

PR #162, opened 2016-07-02
Export a umd bundle and switch to 'index.js'

blink1073:
cc @SylvainCorlay @jasongrout


blink1073:
cc @tkw1536

blink1073:
The last commit makes the output use the `jupyter` namespace, and be available as `jupyter.services`.  I think we should do this for all `jupyter` bundles.

blink1073:
The entry point ends up looking like:

```javascript
(function webpackUniversalModuleDefinition(root, factory) {
	if(typeof exports === 'object' && typeof module === 'object')
		module.exports = factory();
	else if(typeof define === 'function' && define.amd)
		define([], factory);
	else if(typeof exports === 'object')
		exports["services"] = factory();
	else
		root["jupyter"] = root["jupyter"] || {}, root["jupyter"]["services"] = factory();
})
```

blink1073:
cf https://github.com/jupyter/jupyterlab/issues/214

PR #164, opened 2016-07-05
Add mock kernel and session managers and tests

blink1073:
These are needed currently to test the file browser model in JupyterLab.


afshin:
👍

PR #165, opened 2016-07-05
Clean up futures handling in mock kernel

blink1073:
@afshin, now you can write:

```typescript
let kernel = new MockKernel();
kernel.history().then(...);
kernel.sendShellReply(<fake history content>);
```


afshin:
Sweet, thanks!

👍

PR #166, opened 2016-07-05
Kernel message filters

jasongrout:
The IOPub hook system allows you to preempt the handlers for IOPub messages with a
given parent_header message id. 
* The most recently registered hook is run first.
* If the hook returns false, any later hooks and the future's onIOPub handler will not run.
* If a hook throws an error, the error is logged to the console and the next hook is run.
* If a hook is registered during the hook processing, it won't run until the next message.
* If a hook is disposed during the hook processing, it will be deactivated immediately.

### DIFF
--- a/src/ikernel.ts
+++ b/src/ikernel.ts
@@ -286,7 +286,7 @@ interface IKernel extends IDisposable {
    * #### Notes
    * If the callback returns false, the message processing will stop.
    */
-  registerMessageHook(msg_id: string, hook: (msg: KernelMessage.IIOPubMessage) => boolean):  IDisposable;
+  registerMessageHook(msg_id: string, hook: (msg: KernelMessage.IIOPubMessage) => boolean): IDisposable;
 
   /**
    * Get the kernel spec associated with the kernel.
@@ -297,15 +297,6 @@ interface IKernel extends IDisposable {
    * Optional default settings for ajax requests, if applicable.
    */
   ajaxSettings?: IAjaxSettings;
-}
-
-export
-interface IHookList<T> extends IDisposable {
-  add(hook: (msg: T) => boolean): void;
-
-  remove(hook: (msg: T) => boolean): void;
-
-  process(msg: T): boolean;
 }
 
 /**
@@ -428,9 +419,15 @@ namespace IKernel {
     onDone: () => void;
 
     /**
-     * The message hooks
+     * Register hook for IOPub messages.
+     * 
+     * @returns a disposable to unregister the hook.
+     * 
+     * #### Notes
+     * The most recently-registered hook is run first.
      */
-    hooks: IHookList<KernelMessage.IIOPubMessage>;
+    registerMessageHook(hook: (msg: KernelMessage.IIOPubMessage) => boolean): void;
+    removeMessageHook(hook: (msg: KernelMessage.IIOPubMessage) => boolean): void;
   }
 
   /**
--- a/src/ikernel.ts
+++ b/src/ikernel.ts
@@ -275,6 +275,15 @@ interface IKernel extends IDisposable {
   registerCommTarget(targetName: string, callback: (comm: IKernel.IComm, msg: KernelMessage.ICommOpenMsg) => void): IDisposable;
 
   /**
+   * Register an iopub message hook.
+   *
+   * TODO: we could also return a disposable to remove the message hook, but that would mean that you have to store it to be able to delete it. For some reason, I thought it would be better to have an explicit way to deregister a hook.
+   */
+  registerMessageHook(msg_id: string, hook: (msg: KernelMessage.IIOPubMessage) => boolean): void;
+  removeMessageHook(msg_id: string, hook: (msg: KernelMessage.IIOPubMessage) => boolean): void;
+
+
+  /**
    * Get the kernel spec associated with the kernel.
    */
   getKernelSpec(): Promise<IKernel.ISpec>;
@@ -285,6 +294,14 @@ interface IKernel extends IDisposable {
   ajaxSettings?: IAjaxSettings;
 }
 
+export
+interface IHookList<T> extends IDisposable {
+  add(hook: (msg: T) => boolean): void;
+
+  remove(hook: (msg: T) => boolean): void;
+
+  process(msg: T): boolean;
+}
 
 /**
  * A namespace for kernel types, interfaces, and type checker functions.
@@ -404,6 +421,11 @@ namespace IKernel {
      * The done handler for the kernel future.
      */
     onDone: () => void;
+
+    /**
+     * The message hooks
+     */
+    hooks: IHookList<KernelMessage.IIOPubMessage>;
   }
 
   /**
--- a/src/ikernel.ts
+++ b/src/ikernel.ts
@@ -420,13 +420,27 @@ namespace IKernel {
 
     /**
      * Register hook for IOPub messages.
-     * 
-     * @returns a disposable to unregister the hook.
-     * 
+     *
+     * @param hook - The callback invoked for an IOPub message.
+     *
      * #### Notes
-     * The most recently-registered hook is run first.
+     * The IOPub hook system allows you to preempt the handlers for IOPub messages handled
+     * by the future. The most recently registered hook is run first.
+     * If the hook returns false, any later hooks and the future's onIOPub handler will not run.
+     * If a hook throws an error, the error is logged to the console and the next hook is run.
+     * If a hook is registered during the hook processing, it won't run until the next message.
+     * If a hook is removed during the hook processing, it will be deactivated immediately.
      */
     registerMessageHook(hook: (msg: KernelMessage.IIOPubMessage) => boolean): void;
+
+    /**
+     * Remove a hook for IOPub messages.
+     *
+     * @param hook - The hook to remove.
+     *
+     * #### Notes
+     * If a hook is removed during the hook processing, it will be deactivated immediately.
+     */
     removeMessageHook(hook: (msg: KernelMessage.IIOPubMessage) => boolean): void;
   }
 
--- a/src/ikernel.ts
+++ b/src/ikernel.ts
@@ -275,13 +275,18 @@ interface IKernel extends IDisposable {
   registerCommTarget(targetName: string, callback: (comm: IKernel.IComm, msg: KernelMessage.ICommOpenMsg) => void): IDisposable;
 
   /**
-   * Register an iopub message hook.
+   * Register an IOPub message hook.
    *
-   * TODO: we could also return a disposable to remove the message hook, but that would mean that you have to store it to be able to delete it. For some reason, I thought it would be better to have an explicit way to deregister a hook.
+   * @param msg_id - The message id the hook will intercept.
+   *
+   * @param hook - The callback invoked for the message.
+   *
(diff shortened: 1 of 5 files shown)


jasongrout:
A couple of design considerations:

1. We try to defensively architect things so that a consumer that registers a hook does not retain a reference to the future. At the kernel future level, this means having separate registration/removal functions instead of a single registration function returning a disposable. At the kernel level, it means just retaining the msg_id rather than a reference to the future, and looking up the future in the disposable returned from registration.

2. At one point I tried making the registration hook list an attached property of the future assigned by the kernel. But since the hook is most naturally called by the future, we put it back to being a private attribute of the future.

3. At one point I exposed the hook list in the future. However, that also exposes the hook processing function. I opted instead to expose a registration/removal function on the future to hide implementation details.

jasongrout:
I'm not completely happy with the registration/removal function names in the kernel future.

jasongrout:
Also, stdin messages are not handled currently. We could handle then by modifying the typing to include those types of messages, and then inserting the processing in similar to the processing for iopub messages.

blink1073:
You can just have the future itself return a disposable, and make it part of the `IFuture` interface.

blink1073:
If you run `execute`, you have a handle on the future.

jasongrout:
Also, the hook list logic was inspired a lot by phosphor's similar logic in various places and conversations with Chris Colbert. Thanks @sccolbert!

jasongrout:
@blink1073 - the common case here, I think, will be some outside consumer that *doesn't* have a reference to the future asking the kernel to install a hook for some message id. That's how I'm thinking about the output widget, for example. If you already have a reference to the future, you can just directly modify the message processing functions, and you don't need to use the hook system. That's why I reversed an earlier design that had the future returning a disposable (and thus a reference to the future).

jasongrout:
The point of this hook system is to preempt the normal future message processing, which wouldn't be necessary if you already had a reference to the future...

blink1073:
How would you have the msg_id but not the future?

blink1073:
If you want to make it private, I'd say put it on `KernelFutureHandler` but have it return a disposable to the kernel and have it be a pass-through.

jasongrout:
You'd have the msg_id in a message handler. In the output widget's case, you'd have the message id from a comm message.

> If you want to make it private, I'd say put it on KernelFutureHandler but have it return a disposable to the kernel and have it be a pass-through.

I'm not sure what you mean by this. Do you mean have the kernel future return a DisposableDelegate that removes the hook? This would still retain a reference to the future's hook list, right? We could keep it from retaining a reference to the future itself, probably by just using a `function() {}` style function.

src/ikernel.ts
```diff
@@ -404,6 +417,17 @@ namespace IKernel {
...
+
+    /**
+     * Register hook for IOPub messages.
+     * 
+     * @returns a disposable to unregister the hook.
+     * 
+     * #### Notes
+     * The most recently-registered hook is run first.
+     */
+    registerMessageHook(hook: (msg: KernelMessage.IIOPubMessage) => boolean): void;
```
blink1073: This could return an IDisposable, then registerMessageHook on the kernel would just return this value.

jasongrout: Right. That was my iteration before this one. The problem again is that the thing registering the hook then holds a reference to the hook list, at least, and naively would also hold a reference to the future itself. That's what I was hoping avoid. In the end, I decided that registering a hook on the future was probably a relatively rare thing, since you could just modify the message handler directly. I kept the disposable on the kernel to still have a nice api there, but the disposable returned there doesn't retain a reference to the future or hook list.

blink1073: I see, what you are saying, by returning a disposable delegate with a closure, we are keeping a reference in memory to that object.  From an API standpoint, I think it would make more sense to align the methods then so that you use `register/remove` on both the kernel and the future.

jasongrout: Good point about consistency. My hope was that most people would just use the kernel, so it was a concession to convenience for them, and it made the kernel more consistent (e.g., the comm registration returns a disposal). Okay, I'm figuring out your testing framework (looks really nice!), and then I'll start using it and make a choice between keeping the kernel as disposal returning or as register/remove functions. Thanks for your feedback!

jasongrout:
@blink1073 - I think this is ready for review. I think we should support a disposable interface at the kernel level, since that is consistent with the comm target functionality.

jasongrout:
@sccolbert - you might be interested in looking at this too.

src/ikernel.ts
```diff
@@ -275,6 +275,20 @@ interface IKernel extends IDisposable {
   registerCommTarget(targetName: string, callback: (comm: IKernel.IComm, msg: KernelMessage.ICommOpenMsg) => void): IDisposable;
 
   /**
+   * Register an IOPub message hook.
+   *
+   * @param msg_id - The message id the hook will intercept.
```
blink1073: "The message id on which"

src/ikernel.ts
```diff
@@ -275,6 +275,20 @@ interface IKernel extends IDisposable {
...
+   *
+   * @param hook - The callback invoked for the message.
+   *
+   * @returns A disposable used to unregister the message hook
```
blink1073: End with a period.

src/ikernel.ts
```diff
@@ -275,6 +275,20 @@ interface IKernel extends IDisposable {
...
+   *
+   * #### Notes
+   * If the callback returns false, the message processing will stop.
```
blink1073: Refer to `[[IFuture.registerMessageHook]]` since it has more explanation.

src/kernel.ts
```diff
@@ -774,6 +774,36 @@ class Kernel implements IKernel {
...
+   * Register an IOPub message hook.
+   *
+   * @param msg_id - The parent_header message id the hook will intercept.
+   *
+   * @param hook - The callback invoked for the message.
+   *
+   * @returns A disposable used to unregister the message hook.
+   *
+   * #### Notes
+   * The IOPub hook system allows you to preempt the handlers for IOPub messages with a
```
blink1073: Also refer to `IFuture` here.

src/kernelfuture.ts
```diff
@@ -129,25 +160,27 @@ class KernelFutureHandler extends DisposableDelegate implements IKernel.IFuture
...
-    if (this._testFlag(KernelFutureFlag.GotIdle)) {
+    if (reply) { reply(msg); }
+    this._setFlag(Private.KernelFutureFlag.GotReply);
+    if (this._testFlag(Private.KernelFutureFlag.GotIdle)) {
       this._handleDone();
     }
   }
 
   private _handleStdin(msg: KernelMessage.IStdinMessage): void {
+    //let process = this._hooks.process(msg);
```
blink1073: Did you intend to leave this commented out?

src/ikernel.ts
jasongrout: yeah, I agree that's awkward. I'll rephrase.

test/src/testkernel.ts
```diff
@@ -1245,6 +1246,497 @@ describe('jupyter.services - kernel', () => {
 
     });
 
+    context('#registerMessageHook()', () => {
+            it('should have the most recently registered hook run first', (done) => {
```
blink1073: Formatting out of whack.

src/ikernel.ts
jasongrout: Ah, I thought I got all those!

test/src/testkernel.ts
```diff
@@ -1245,6 +1246,497 @@ describe('jupyter.services - kernel', () => {
...
+                done();
+              });
+            };
+
+          });
+        });
+      });
+
+
+      it('should abort processing if a hook returns false, but the done logic should still work', (done) => {
```
blink1073: extra line

blink1073: Mis-nested tests and extra lines below.

src/ikernel.ts
jasongrout: I copied the full explanation here.

src/kernelfuture.ts
jasongrout: I intended to remove it and address stdin messages when we can handle them.

blink1073: Can we remove the comments and open an issue instead?

test/src/testkernel.ts
```diff
@@ -1245,6 +1246,488 @@ describe('jupyter.services - kernel', () => {
...
+              expect(calls).to.eql(['last', 'iopub', 'first', 'last', 'iopub']);
+              doLater(() => {
+                done();
+              });
+            };
+          });
+        });
+      });
+
+      it('removing a message hook in another message hook deactivates it immediately', (done) => {
```
blink1073: Should start with "should"

test/src/testkernel.ts
```diff
@@ -1245,6 +1246,488 @@ describe('jupyter.services - kernel', () => {
...
+      createKernel().then(kernel => {
+        let future = kernel.execute({ code: 'hello' });
+        expect(typeof future.msg.header.msg_id).to.be('string');
+        expect(() => { future.msg = null; }).to.throwError();
+        done();
+      });
+    });
+
+    describe('Message hooks', () => {
+      it('should have the most recently registered hook run first', (done) => {
```
blink1073: Missing line

test/src/testkernel.ts
```diff
@@ -1245,6 +1246,488 @@ describe('jupyter.services - kernel', () => {
...
+              expect(calls).to.eql(['last', 'iopub', 'first', 'last', 'iopub']);
+              doLater(() => {
+                done();
+              });
+            };
+          });
+        });
+      });
+
+      it('removing a message hook in another message hook deactivates it immediately', (done) => {
```
blink1073: Missing a should

test/src/testkernel.ts
jasongrout: where should the new line go?

jasongrout: Never mind, I now see that the pattern in the file is to put a newline after describe/context blocks.

jasongrout:
@blink1073 - I rebased my last commit to include fixes for your last few comments. Thanks!

jasongrout:
When this goes in, I'll open an issue to handle stdin requests as well as iopub requests.

src/kernel.ts
jasongrout: Fixed.

blink1073:
Is this ready to go?  If so, can you update the top level comment?

test/src/testkernel.ts
```diff
@@ -1245,6 +1246,491 @@ describe('jupyter.services - kernel', () => {
...
+            future.onIOPub = () => {
+              calls.push('iopub')
+            };
+
+            future.onDone = () => {
+              expect(calls).to.eql(['first', 'delete', 'iopub', 'first', 'iopub']);
+              doLater(() => {
+                done();
+              });
+            };
```
blink1073: Just as a note, we typically add a line between the closing of the test functions as well.

jasongrout: Fixed

blink1073:
Looks good, thanks!  I'll cut a release after #178 is merged as well.

blink1073:
Released in v0.13.0.

jasongrout:
Awesome, thanks!

PR #170, opened 2016-07-06
Add Terminals services

blink1073:
Since it is using the server apis, it made sense to move this here from `JupyterLab`.


minrk:
Makes sense, thanks.

PR #171, opened 2016-07-07
add browser-require example

minrk:
uses ES5, require to load webpack bundles of services/utils

still installed with npm.

TODO:

- [x] need release of jupyter-js-utils with webpack bundle so that it doesn't need to come from a local file.


minrk:
Alternate, possibly preferable: could skip package.json entirely, and get utils and services from CDN. That would require releases of both utils and services.

blink1073:
We still have the larger question of whether these dist bundles should be self-contained or not. cf https://github.com/jupyter/jupyterlab/issues/214.

minrk:
Right.

blink1073:
To me it seems too arduous to force a user to manually include all of the dependencies.

blink1073:
This is different from the way it would be handled in JupyterLab, where we would explicitly be providing the individual bundles.

minrk:
Right, this is only relevant for those not doing full webpack bundles. In general, *our* applications should use bundles to produce as close to a single .js file per page or application (excluding extensions) as we can get.

All dependencies definitely shouldn't be necessary. In this case, it should be *at most* services and utils, but possibly just services. And services should probably work with nothing else loaded, as long as *I* don't use utils.  I'm less sure about widgets bundling services or not. Maybe we should produce both thin and fat bundles?

blink1073:
Even if you did use utils, the version used by the services bundle would be self-contained.  It is definitely a balancing act.

blink1073:
https://npmcdn.com/jupyter-js-services@0.12.0/dist/ is now available.

minrk:
And have you published utils, as well?

blink1073:
The services bundle includes utils for now.

SylvainCorlay:
What do you recommend doing in jupyter-js-widgets? 
 - in the case of the bundle for the legacy notebook, we keep it in since it is not part of the notebook.
 - in the case of jupyter-lab, should it be an external?
 - in the case of the cdn bundle, should we grab this one from the cdn too and make it external?

blink1073:
I don't think one webpack bundle can require another one from a cdn link, you'd have to use scriptjs/requirejs and make the import async

blink1073:
As in, the public path only works *within* a package.

SylvainCorlay:
Actually, you can! 

For example, in the bqplot bundle, we export an amd module, with jupyter-js-widgets as an external, and the bundle [here](https://npmcdn.com/bqplot@0.1.1/dist/index.js) looks like

`define(["jupyter-js-widgets"], function(__WEBPACK_EXTERNAL_MODULE_6__)`

SylvainCorlay:
Fetching `jupyter-js-widgets` from npm cdn would only require to add a requirejs-config for the npmcdn module...

blink1073:
Yeah, that is what I said ;)

blink1073:
As in, you cannot do it with webpack alone, you need either scriptjs or requirejs (and we need to document it).

examples/browser-require/index.html
```diff
@@ -0,0 +1,39 @@
...
+    </div>
+    <textarea id="cell" cols="100" rows="24"></textarea>
+    <div>
+      <button id="run">run</button>
+    </div>
+    <div id="output"></div>
+    <script type="text/javascript">
+    require.config({
+        paths: {
+          'jquery': 'https://code.jquery.com/jquery-2.2.4.min',
```
blink1073: Does this use the same jquery as the script tag above?

minrk:
This example now loads jupyter-js-services from npmcdn with no package.json or local dependencies.

SylvainCorlay:
> Yeah that is what I said ;)

ok, I was just revealing the deeper meaning of your word to the common people :)

examples/browser-require/index.html
minrk: You caught me in the middle of adding this to replace the script tag above. The script tag is now removed.

examples/browser-require/index.html
```diff
@@ -0,0 +1,38 @@
...
+          'jupyter-js-services': 'https://npmcdn.com/jupyter-js-services@0.12.0/dist/index',
```
SylvainCorlay: You can add a `^` in the semver string!

SylvainCorlay:
y9tcz

SylvainCorlay:
@blink1073 why does https://npmcdn.com/jupyter-js-services@0.12.0/dist/ also contain a jupyter-js-services bundle?

examples/browser-require/index.html
SylvainCorlay: Besides, why not `https://npmcdn.com/jquery@3.0.0/dist/jquery.min.js`? (or another version?)

blink1073:
Cool, thanks @minrk!

minrk:
Thanks!​

PR #173, opened 2016-07-07
Fix disposal handling in mock kernel

blink1073:
This was causing a failure in jupyterlab when trying to upgrade this dependency. cf https://github.com/jupyter/jupyterlab/pull/245


afshin:
👍

PR #176, opened 2016-07-07
Minor cleanup of browser-require example

blink1073:
cc @minrk 

The kernel info console log was incorrect, and a mis-spelling of `use strict`.  Also made the string quoting consistent.


willingc:
LGTM.

PR #179, opened 2016-07-08
Add a session manager and make models json objects

blink1073:
This functionality was previously in JupyterLab but belongs here.


afshin:
👍

PR #188, opened 2016-07-11
Add signals for changes affecting the managers

### LINKED ISSUE #187: Add signals to the managers to trigger updates
Creating or shutting down should also call `list()`, and a call to `list()` should result in a signal that others can listen to.

We should also support updating the kernelspecs from the services manager.

blink1073:
Fixes #187.

### DIFF
--- a/src/kernel.ts
+++ b/src/kernel.ts
@@ -22,7 +22,7 @@ import {
 } from './ikernel';
 
 import {
-  JSONObject, JSONValue
+  JSONObject, JSONValue, deepEqual
 } from './json';
 
 import {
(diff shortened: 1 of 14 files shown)


src/kernel.ts
```diff
@@ -76,13 +118,23 @@ class KernelManager implements IKernel.IManager {
    * @param options - Overrides for the default options.
    */
   listRunning(options?: IKernel.IOptions): Promise<IKernel.IModel[]> {
-    return listRunningKernels(this._getOptions(options));
+    return listRunningKernels(this._getOptions(options)).then(running => {
+      if (!deepEqual(running, this._running)) {
+        this.runningChanged.emit(running);
+        this._running = running.slice();
```
afshin: I'd swap the order of these two lines.

src/session.ts
```diff
@@ -50,12 +54,53 @@ class SessionManager implements ISession.IManager {
...
+  /**
    * Get the available kernel specs. See also [[getKernelSpecs]].
    *
    * @param options - Overrides for the default options.
    */
   getSpecs(options?: ISession.IOptions): Promise<IKernel.ISpecModels> {
-    return getKernelSpecs(this._getOptions(options));
+    return getKernelSpecs(this._getOptions(options)).then(specs => {
+      if (!deepEqual(specs, this._specs)) {
+        this.specsChanged.emit(specs);
```
afshin: `this._specs` should be set here.

src/session.ts
```diff
@@ -64,14 +109,24 @@ class SessionManager implements ISession.IManager {
    * @param options - Overrides for the default options.
    */
   listRunning(options?: ISession.IOptions): Promise<ISession.IModel[]> {
-    return listRunningSessions(this._getOptions(options));
+    return listRunningSessions(this._getOptions(options)).then(running => {
+      if (!deepEqual(running, this._running)) {
+        this.runningChanged.emit(running);
+        this._running = running.slice();
```
afshin: I'd swap these two as well.

afshin:
👍

PR #192, opened 2016-07-13
Make hooks robust against disposal

jasongrout:
There were a few corner cases where trying to remove a hook on a disposed future could cause problems.


blink1073:
The best offense is a good defense.

PR #193, opened 2016-07-16
Make the kernelspecs fetch more fault tolerant

blink1073:
cc @afshin


afshin:
👍 

Thanks!

damianavila:
This is not directly impacting jupyterlab now, but just to let you know, we've released versions of nb_conda, nb_conda_kernels and nb_anacondacloud through the official Anaconda conda repos that should resolve the issues you saw with Python[root].

PR #194, opened 2016-07-18
Add a url lookup function for files

blink1073:

cc @jasongrout @minrk 

Uses the `node` path functions.

Exposes the path manipulation methods as static members of `ContentsManager` for general use.


src/contents.ts
```diff
@@ -353,6 +379,42 @@ class ContentsManager implements IContents.IManager {
...
+   *
+   * #### Notes
+   * If the path is not contained within the server path, it will be
+   * returned unmodified.
+   * If the relativePath starts with a forward slash, the basePath will
+   * be ignored and a path relative to the server root will be returned.
+   */
+  getPath(relativePath: string, basePath = ''): string {
+    let norm = pathutil.normalize(pathutil.join(basePath, relativePath));
+    if (norm.indexOf('..' + pathutil.sep) === 0) {
```
minrk: In js-services, all paths will be URLs joined by `/`, so `path.sep` would do the wrong thing on Windows, wouldn't it?

blink1073: Yeah, I had the hair-brained idea to make this "cross platform" last night, fixing now ;).

blink1073:
This is ready.

jasongrout:
Thinking more about this, I'm afraid using an os path library will not work. For example, what if the path name has a `?` in it? That's a reserved character in URLs, but not in paths. URLs have different rules than posix paths, so we should be using a URL library that will properly escape things too.

jasongrout:
It looks like https://www.npmjs.com/package/url is the package we want.

jasongrout:
Never mind, I was being dumb. The path resolution should happen in posix space, and *then* converted to a URL. Otherwise question marks in paths, for example, will be misinterpreted.

However, as a separate point, perhaps the utils should be using the url package rather than writing its own functions to do things with urls.

blink1073:
Sounds good.  How about we merge this, then #195, then I'll add the `url` lib dependency?

src/contents.ts
```diff
@@ -656,4 +675,59 @@ namespace ContentsManager {
...
     ajaxSettings?: IAjaxSettings;
   }
+
+  /**
+   * Get a path to a file given a relative path.
+   *
+   * @param relativePath - The relative path to the file.
+   *
+   * @param basePath - The optional base path of the file.  The
+   *   default is the server root path.
```
jasongrout: `The default is ''.`

src/contents.ts
```diff
@@ -656,4 +675,59 @@ namespace ContentsManager {
...
+   *
+   * #### Notes
+   * If the path is not contained within the server path, it will be
```
jasongrout: with the *base* path

blink1073: Both updated.

src/contents.ts
```diff
@@ -656,4 +675,59 @@ namespace ContentsManager {
...
+   * returned unmodified.
+   * If the relativePath starts with a forward slash, the basePath will
```
jasongrout: This function is really confusing. If relativePath is actually an absolute path, then ignore base path. But if it is relative, then ensure that base path is included, unless the relative path actually goes above base path, in which case we just return the relative path, which may still go above the base path, courtesy of `../` components.

blink1073: I removed the comment about the forward slash, since that is the default behavior of `resolve` anyway.  I'll try and clean up the language some more.

blink1073: Updated.

blink1073:
I updated the description and behavior of `getAbsolutePath` based on discussion with @afshin, how does that look to you, @jasongrout?

test/src/testcontents.ts
```diff
@@ -103,6 +102,64 @@ describe('jupyter.services - Contents', () => {
...
+    it('should yield the current directory', () => {
+      let path = ContentsManager.getAbsolutePath('.', 'fizz/buzz');
+      expect(path).to.be('fizz/buzz');
+    });
+
+    it('should yield the parent directory', () => {
+      let path = ContentsManager.getAbsolutePath('..', 'fizz/buzz');
+      expect(path).to.be('fizz');
+    });
+
```
jasongrout: Can we make a test showing that things are *not* URL encoded? Perhaps a path with a space and a question mark in it?

test/src/testcontents.ts
```diff
@@ -103,6 +102,64 @@ describe('jupyter.services - Contents', () => {
...
+    it('should get the url of a file', () => {
+      let contents = new ContentsManager({ baseUrl: 'http://foo', });
+      let url = contents.getDownloadUrl('bar.txt');
+      expect(url).to.be('http://foo/files/bar.txt');
+      url = contents.getDownloadUrl('fizz/buzz/bar.txt');
+      expect(url).to.be('http://foo/files/fizz/buzz/bar.txt');
+      url = contents.getDownloadUrl('/bar.txt');
+      expect(url).to.be('http://foo/files/bar.txt');
+    });
```
jasongrout: Can we make a test showing that things *are* URL-encoded? Perhaps a path with a space and a question mark in it?

jasongrout: Maybe one more test for `url = contents.getDownloadUrl('../bar.txt');`

blink1073: As in, 

```typescript
    it('should handle a relative path', () => {
      let contents = new ContentsManager({ baseUrl: 'http://foo', });
      let url = contents.getDownloadUrl('../bar.txt');
      expect(url).to.be('http://foo/files/../bar.txt');
    });
```

blink1073: `getDownloadUrl` does not do any resolution, it just does url encode + join.

test/src/testcontents.ts
```diff
@@ -158,6 +163,18 @@ describe('jupyter.services - Contents', () => {
...
+    it('should encode a question mark', () => {
+      let contents = new ContentsManager({ baseUrl: 'http://foo', });
+      let url = contents.getDownloadUrl('bar?3.txt');
+      expect(url).to.be('http://foo/files/bar%3F3.txt');
+    });
+
+    it('should handle a relative path', () => {
+      let contents = new ContentsManager({ baseUrl: 'http://foo', });
+      let url = contents.getDownloadUrl('../bar.txt');
+      expect(url).to.be('http://foo/files/../bar.txt');
```
jasongrout: I thought it should return null or something. What do you think?

jasongrout: Oh, you're saying the resolution should happen strictly before this?

jasongrout: Okay, +1

jasongrout:
Looks good now. +1!

test/src/testcontents.ts
blink1073: No, I think it is fair to allow a two-for-one in this case.  Updating.

src/contents.ts
```diff
@@ -353,6 +365,13 @@ class ContentsManager implements IContents.IManager {
   }
 
   /**
+   * Get a download url given a file path.
```
jasongrout: We should clarify that the path is a posix file path, *not* URL-encoded. Perhaps in a `@param` listing?

blink1073: Updated to separate concerns, and updated documentation.

src/contents.ts
```diff
@@ -186,7 +186,9 @@ namespace IContents {
     get(path: string, options?: IFetchOptions): Promise<IModel>;
 
     /**
-     * Get a download url given an absolute file path.
+     * Get a an encoded download url given a file path.
```
jasongrout: Get an encoded.. (delete 'a')

src/contents.ts
```diff
@@ -365,7 +365,15 @@ class ContentsManager implements IContents.IManager {
   }
 
   /**
-   * Get a download url given a file path.
+   * Get a an encoded download url given a file path.
```
jasongrout: Get an encoded.. (delete 'a')

jasongrout:
Thanks, the updated docs look great (modulo a typo).

jasongrout:
:shipit:

blink1073:
Shipping.

Issue #190, opened 2016-07-11
Reorganize the repository


blink1073:
Current plan:

```
services/src/
   kernel/
      index.ts - Kernel n/s, IKernel, KernelManager, KernelMessage n/s, 
      kernel.ts
      comm.ts - not exported
      future.ts - not exported
      messages.ts - KernelMessage n/s
      serialize.ts
      manager.ts
      validate.ts
   session/
      index.ts - Session n/s, ISession, SessionManager
      session.ts
      manager.ts
      validate.ts
   terminal/
      index.ts - Terminal n/s, ITerminal, TerminalManager
      terminal.ts
      manager.ts
      validate.ts
   contents/
     index.ts - Contents n/s, IContentsManager, ContentsManager
     contents.ts
     manager.ts
     validate.ts
   config/
     index.ts - Config n/s, IConfigSection, 
     config.ts
     validate.ts
   common/
      index.ts
   manager.ts
   mockserver.ts
   index.ts
```

- The namespaces have interfaces, helper functions and factory functions
- The implementations of Kernel, Comm, Future, Session, and Terminal 
should be private and use a different name.

blink1073:
Completed by the above referenced PRs.

Issue #189, opened 2016-07-11
Switch to a mock server instead of mock client side objects


blink1073:
We can use `ws` as both the `WebSocket` implementation and server, and `sinon` to mock XHR.

Use https://github.com/blink1073/jupyter-js-services/blob/f31bd4bf614aef543bac858b32e7bcbe5b24ebcc/src/mockserver.ts as a starting point.

PR #196, opened 2016-07-20
[WIP] Connection and mock overhaul

blink1073:
- Expose WebSocket connection uniformly between the kernel and the terminal, with a connection state and signal.  
- Kernels and terminals no longer connect to the websocket automatically.  
- Expose the http and websocket urls so we can use them for mock-xhr and mock-socket.  
- Use the mock libraries instead of the custom ones.
- Create utility functions that can be used to mock responses from the server, and use these instead of the mock client/server objects.


blink1073:
I'm going to consider this a proof of concept, informing #190, since I think it should be done at the same time.

blink1073:
But I'm going to wait until after we switch to the mono phosphor repo.

blink1073:
cf #189

PR #202, opened 2016-07-23
Clean up utils and use external url functions

blink1073:
I could not find a suitable ajax library that has the features we are using.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
     "path-posix": "^1.0.0",
     "phosphor-disposable": "^1.0.5",
     "phosphor-signaling": "^1.2.0",
-    "requirejs": "^2.2.0"
+    "requirejs": "^2.2.0",
+    "url": "^0.11.0"
   },
   "devDependencies": {
     "expect.js": "^0.3.1",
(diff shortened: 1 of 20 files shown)


package.json
```diff
@@ -9,7 +9,8 @@
     "path-posix": "^1.0.0",
     "phosphor-disposable": "^1.0.5",
     "phosphor-signaling": "^1.2.0",
-    "requirejs": "^2.2.0"
+    "requirejs": "^2.2.0",
+    "url": "^0.11.0"
```
jasongrout: FYI, I noticed that both url and the path module we are using are missing upstream bugfixes from npm (cf. https://github.com/defunctzombie/node-url/issues/24). They may be better than if we tried to write something from scratch, though...

blink1073: I'm actually not using `url`, it should have been `url-join`, since `url` didn't offer anything over `posix-path`.  Agreed that we don't want to have our own path implementation.

blink1073:
Note: I am going to use `url.resolve` for of `urlPathJoin` and `urlJoinEncode`, since it handles both cases nicely.

blink1073:
Ready.

afshin:
👍

PR #203, opened 2016-07-25
Add cwd to the contents manager

blink1073:

This provides a common `cwd` to all consumers of a contents manager.
This will replace the `PathTracker` in the application.

### DIFF
--- a/src/contents.ts
+++ b/src/contents.ts
@@ -1,8 +1,13 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
+
 import {
-  IAjaxSettings
-} from './utils';
+  IDisposable
+} from 'phosphor-disposable';
+
+import {
+  ISignal, Signal, clearSignalData
+} from 'phosphor-signaling';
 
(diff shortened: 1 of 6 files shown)


minrk:
I'm not entirely sure what `cwd` means for the ContentsManager. Shouldn't it be stateless and never change?

blink1073:
I think it makes sense to have a convenience cwd on the client-side for creating new objects and resolving paths, plus this object is already shared by most of the application.

minrk:
It seems like this is the wrong place for that state, though. Shouldn't it be state of the Application using the contents service, and not state on the contents service itself?

blink1073:
I had originally attached this to the `ServiceManager` itself, which aggregates all of the services and is what is passed around in the application.  Would you be more comfortable putting it there?

blink1073:
I am starting to see what you mean though, this should be in an `ApplicationState` object (that we don't yet have).

blink1073:
Okay, I'll close this then.

minrk:
That makes sense, thanks!

Issue #205, opened 2016-07-25
With credentials usage


hamsal:
The documentation seems sparse which I understand is because the project is so young, I can't seem to find anything about creating a session pointing to jupyter server that has password auth enabled. The code snippet I am trying to work it into looks like this, trying to use "withCredentials":

```
var options = {
  baseUrl: BASE_URL,
  wsUrl: WS_URL,
  kernelName: 'python',
  path: '/tmp/foo.ipynb'
}
services.startNewSession(options).then(function(session) {
    // Execute and handle replies on the kernel.
    var future = session.kernel.execute({ code: code });
    future.onReply = function(reply) {
 ...
```

I took a guess and tried:

```
var options = {
  baseUrl: BASE_URL,
  wsUrl: WS_URL,
  kernelName: 'python',
  path: '/tmp/foo.ipynb'
  withCredentials: true,
  user: 'foo',
  password: 'bar'
}
```

blink1073:
Hi @hamsal, we could certainly use an example with this behavior.  As for your direct question, your last three attributes should all be under `ajaxSettings`:

```typescript
var options = {
  baseUrl: BASE_URL,
  wsUrl: WS_URL,
  kernelName: 'python',
  path: '/tmp/foo.ipynb',
  ajaxSettings: {
    withCredentials: true,
    user: 'foo',
    password: 'bar'
 }
}
```

hamsal:
Thank you! I just pulled to most recent master and ran your corrected version. What are some steps I could take to troubleshoot the 403 forbidden code: 

On client side I get:
```API request failed (403):  null``` 

On the jupyter server logs I see 
```[W 00:38:26.154 NotebookApp] 403 POST /dummy_user/api/sessions?1469493506125 (192.168.99.1) 1.73ms referer=None```

I am using the same password in place of 'bar' that I have configured in my jupyter server and the password works when I login to the jupyter server using a browser.

**Comment by blink1073** [comment]
Can you run the notebook server with a `--debug` flag?

**Comment by hamsal** [comment]
Sure, Here are the potentially relevant bits (Removing unrelated fluff from an unrelated process I run)

```
[D 00:59:48.490 NotebookApp] No files in /root/.ipython/nbextensions
[D 00:59:48.492 NotebookApp] Searching ['/user', '/root/.jupyter', '/usr/etc/jupyter', '/usr/local/etc/jupyter', '/etc/jupyter'] for config files
[...]

PR #206, opened 2016-07-26
[WIP] Secure browser example

blink1073:
I followed the instructions as best as I can, but `@web.authenticated` requires the user to be actually logged in for anything but `GET` and `HEAD` requests, and I can't figure out how to actually perform the login.

"Failed to load resource: the server responded with a status of 403 (Forbidden)"

http://jupyter-notebook.readthedocs.io/en/latest/public_server.html

@minrk, lifeline?

cf #205

### DIFF
--- a/examples/browser-secure/index.js
+++ b/examples/browser-secure/index.js
@@ -3,21 +3,24 @@ require(['jquery', 'jupyter-js-services'], function ($, services) {
   'use strict';
 
   var baseUrl = services.utils.getBaseUrl();
-  var url = services.utils.urlPathJoin(baseUrl, 'login');
-  services.utils.ajaxRequest(url, {
-    data: JSON.stringify({ password: 'password' }),
-    method: 'POST',
-    contentType: 'application/json'
-  }).then(function (success) {
+
+  // need ajaxSettings.withCredentials = true
+  // for cookies to be sent to the server
+  var options = {
+    ajaxSettings: {
+      withCredentials: true,
+    }
+  };
+
+  services.listRunningKernels(options)
+    .then(function (success) {
 
       var startNewKernel = services.startNewKernel;
+      $('#kernel-info').text('Starting...');
 
       var kernelOptions = {
         name: 'python',
-        ajaxSettings: {
-          withCredentials: true,
-          password: 'password'
-        }
+        ajaxSettings: options.ajaxSettings,
       };
 
     // start a single kernel for the page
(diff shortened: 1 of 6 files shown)


examples/browser-secure/main.py
```diff
@@ -0,0 +1,94 @@
...
+        return self.render("index.html", static=self.static_url,
+                           base_url=self.base_url)
+
+
+def main():
+    # Start a notebook server with cross-origin access.
+    url = "http://localhost:%s" % PORT
+
+    nb_command = [sys.executable, '-m', 'notebook', '--no-browser', '--debug',
+                  '--NotebookApp.password="%s"' % passwd('password'),
```
blink1073: Relevant change is here.

examples/browser-secure/index.js
```diff
@@ -0,0 +1,51 @@
+
+require(['jquery', 'jupyter-js-services'], function ($, services) {
+  'use strict';
+  var startNewKernel = services.startNewKernel;
+
+  var kernelOptions = {
+    name: 'python',
+    ajaxSettings: {
```
blink1073: Other relevant change.

minrk:
For now, a cookie has to be requested. It's uses like this that prompt the desire to add API Token-based authentication (like JupyterHub, Kernel Gateway), which is very simple, but not yet implemented.

To do the cookie auth, you have to POST the login form, and then use those cookies in future requests. With requests in Python:

```python
import requests
base_url = 'http://127.0.0.1:8888'

# make login request to get cookie
login = requests.post(base_url + '/login', data={
        'password': password
    }, allow_redirects=False)
login.raise_for_status()

# pass cookies=login.cookies to authorize future requests
r = requests.get(base_url + '/api/kernels/', cookies=login.cookies)
kernels = r.json()
```

minrk:
In the case of a webapp like this one, you could check for 403, then redirect to the notebook's login page and redirect back here. I can't remember of cookies are sent with cross-origin requests, though.

blink1073:
I futzed with this a bit more.  Sending data on the POST results in an OPTIONS request, which is not supported by the `login/` handler.  I don't think this particular example will work, I'm at a loss.

minrk:
I'll take a stab at it with the redirect. I think browsers probably don't let you do cross-origin login stuff like you would in a Python/nodejs script.

minrk:
I opened https://github.com/blink1073/jupyter-js-services/pull/8 to use the notebook's own login page, and then it works.

blink1073:
I'll fix #208 before finishing this.

minrk:
Cool, thanks!

blink1073:
This is what we get back from the server now:

screen shot 2016-07-29 at 6 19 53 am

blink1073:
Closing as stale.

PR #207, opened 2016-07-27
Add a file extension normalization function and expose url.parse

blink1073:
These will be used in JupyterLab in the document registry and manager.

### DIFF
--- a/src/contents.ts
+++ b/src/contents.ts
@@ -708,6 +708,11 @@ namespace ContentsManager {
    */
   export
   function getAbsolutePath(relativePath: string, cwd = ''): string {
+    // Bail if it looks like a url.
+    let urlObj = utils.urlParse(relativePath);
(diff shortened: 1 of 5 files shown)


src/contents.ts
```diff
@@ -742,4 +745,17 @@ namespace ContentsManager {
...
   }
+
+  /**
+   * Normalize a file extension to be of the type `'.foo'`.
+   *
+   * Adds a leading dot if not present and converts to lower case.
+   */
+  export
+  function normalizeExtension(extension: string): string {
+    if (extension.indexOf('.') !== 0) {
```
minrk: `normalizeExtension('')` should still return `''`, so you may also want `extension.length > 0 &&` here.

blink1073: Done, thanks.

afshin:
👍

PR #210, opened 2016-07-28
Fix handling of request errors

### LINKED ISSUE #208: API request errors should keep a handle on the original .xhr
If an API request fails with 403, 500, etc., it should always be possible to access the status code, and other info attached to the .xhr object. Preserving the .xhr attribute on our re-raised Errors is probably the easiest way to do this.

blink1073:
Fixes #208.

Matches the API of http://api.jquery.com/ajaxSuccess/ and http://api.jquery.com/ajaxError/
All ajax-related methods reject IAjaxError on error.

cc @minrk


blink1073:
Actually, I'm going to make the internal errors throw proper `IAjaxError` as well.

minrk:
Awesome, thanks!

PR #212, opened 2016-07-29
Expose cached kernel info and spec

### LINKED ISSUE #209: cache kernelinfo
In https://github.com/jupyter/jupyter-js-services/blob/fc0e0467c1f573582e7ef9432f351682dbe0b67b/src/kernel.ts#L707, a request is made every time the info is requested. Since it is unlikely to change (and indeed, we don't provide any change signals for it anyway), we should cache it. What about having a new attribute like `.info` that is a promise for the info, and either fetches or resolves to the cached value. That way someone can still fetch new info to refresh the cache, but most users will probably just use the cached value.

cf. https://github.com/jupyter/jupyterlab/issues/520

blink1073:
Fixes #209.

Also use the more targeted method to retrieve the kernel spec.

### DIFF
--- a/src/ikernel.ts
+++ b/src/ikernel.ts
@@ -76,7 +76,7 @@ interface IKernel extends IDisposable {
    * #### Notes
    * This is a read-only property.
    */
-   username: string;
+  username: string;
 
   /**
    * The client unique id.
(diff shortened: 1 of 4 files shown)


afshin:
👍

PR #215, opened 2016-08-30
Use known path in connectToSession

### LINKED ISSUE #214: connectToSession by id does not populate missing session options
I'd like to be able to connect to an existing session using only the session ID and the server URL. The docs suggest that `connectToSession(id, {"baseUrl":"http://localhost:8888"})` should work, which it mostly does. Functionality like running commands works great.

However, the resulting session object is missing certain attributes, for example `path`. I think these attributes should be populated automatically from the server whenever a session is connected, instead of making them mandatory in the call to `connectToSession`.

I can work around this by separately fetching the session model and then calling `session.update`, but I don't believe that this is the intended approach.

blink1073:
Fixes #214.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -752,17 +752,16 @@ namespace Private {
   /**
    * Create a Promise for a kernel object given a session model and options.
    */
-  export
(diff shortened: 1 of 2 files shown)


afshin:
👍

PR #219, opened 2016-09-13
Do not convert extensions to lower case

### LINKED ISSUE #217: File extension only lowercase
I have a question regarding the ContentsManager#normalizeExtension in /jupyter-js-services/src/contents.ts 

What is the reason to lowercase the file extension?

There are languages where the Upper case is also common as a default file extension. IMHO it should be API-user decision that provides the extension for ICreateOptions

blink1073:
Fixes #217.


afshin:
👍

PR #220, opened 2016-09-13
Fix bug in ConfigWithDefaults get for falsey data

### LINKED ISSUE #216: Fix bug in 'ConfigWithDefaults.get'
cf https://github.com/jupyter/notebook/pull/1709, we have the same issue.

blink1073:
Fixes #216.

### DIFF
--- a/test/src/testconfig.ts
+++ b/test/src/testconfig.ts
@@ -179,14 +179,14 @@ describe('jupyter.services - ConfigWithDefaults', () => {
 
     it('should get a falsey value', (done) => {
       let handler = new RequestHandler(() => {
-        handler.respond(200, { testclass: { foo: false } });
+        handler.respond(200, { testclass: { foo: 0 } });
       });
       let defaults: JSONObject = { foo: true };
       let className = 'testclass';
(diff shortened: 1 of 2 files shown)


test/src/testconfig.ts
```diff
@@ -177,6 +177,19 @@ describe('jupyter.services - ConfigWithDefaults', () => {
...
+    it('should get a falsey value', (done) => {
+      let handler = new RequestHandler(() => {
+        handler.respond(200, { testclass: { foo: false } });
+      });
+      let defaults: JSONObject = { foo: true };
+      let className = 'testclass';
+      getConfigSection({ name: 'test' }).then(section => {
+        let config = new ConfigWithDefaults({ section, defaults, className });
+        let data = config.get('foo');
+        expect(data).to.be(false);
```
afshin: Should this be *falsey* or outright `false`? You could use `.to.not.be.ok();' if it's the former.

blink1073: Fixed, thanks.

afshin:
Looks good, thanks!

👍

PR #221, opened 2016-09-13
Refactor Kernel

blink1073:
cf #190.

Moves kernel-related files and code to `src/kernel`.

API changes: 

- Use a `Kernel` namespace instead of the `IKernel` namespace.
- Rename `Kernel` class -> `DefaultKernel`.  
- Move the free kernel-related functions onto the `Kernel` and `KernelMessage` namespaces as appropriate.
- Kernel now exports: `IKernel` interface, `Kernel` namespace, `KernelMessage` namespace, `KernelManager` class. 
- Rename `IKernel.getKernelSpecs` -> `IKernel.getSpecs`


src/kernel/default.ts
```diff
@@ -1267,231 +916,270 @@ class Kernel implements IKernel {
...
+namespace DefaultKernel {
   /**
-   * Construct a new comm channel.
+   * Find a kernel by id.
+   *
+   * #### Notes
+   * If the kernel was already started via `startNewKernel`, we return its
+   * `Kernel.IModel`.
+   *
+   * Otherwise, if `options` are given, we attempt to find to the existing
```
afshin: >  to find ~~to~~ the

src/kernel/index.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+'use strict';
```
afshin: Superfluous `use strict`.

src/kernel/kernel.ts
```diff
@@ -0,0 +1,738 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+'use strict';
```
afshin: Superfluous `use strict`.

src/kernel/index.ts
afshin: There are a bunch of these, I won't bombard the PR marking all of them.

src/kernel/default.ts
```diff
@@ -610,7 +259,7 @@ class Kernel implements IKernel {
    *
    * If the kernel status is `Dead`, this will throw an error.
```
afshin: In the doc strings it says `Dead` but in the code it's `dead` ... I don't know if this makes a big difference, but we may want to sync those up.

src/kernel/index.ts
blink1073: I updated all of the ones in the `kernel` folders.

blink1073:
Comments addressed.

afshin:
Awesome, looks good to me.

👍

PR #222, opened 2016-09-16
Session refactor

blink1073:
cf #190 

- Creates `src/session` folder
- Moves free functions onto `Session` namespace and removes `ISession` namespace
- Renames `Session` -> `DefaultSession`


afshin:
Thanks!

PR #227, opened 2016-09-28
Remove mock socket from tests

blink1073:
Use the `WebSocket` and `Server` from `ws` for testing.

Also refactors the tests to use `beforeEach`/`afterEach` where appropriate.

We still use the custom mock xhr for now because `sinon`'s typings file does not include a callback for when a mock xhr object is sent nor an event listener interface.


afshin:
Well hot diggity.

🌶

PR #230, opened 2016-10-03
Use a temporary directory for the integration test

### LINKED ISSUE #229: Provide a server extension instead of mock server objects
- We can use a server extension that provides a contents manager that uses a mock file system, similar to the [labextension tests](https://github.com/jupyter/jupyterlab/blob/master/jupyterlab/tests/test_labextensions.py).  
- We could add a few other custom capabilities as needed.
- This would make it transparent to test objects that use jupyter-js-services.
- We would use a similar script to the [integration test](https://github.com/jupyter/jupyter-js-services/blob/master/test/integration_test.py) to bootstrap the Jupyter server and the karma server.
- The local unit tests would still use `ws` and `sinon` for testing, because we need to simulate bad responses from the server.

@jasongrout, @minrk, thoughts on this approach?

blink1073:
Fixes #229.

A variant of this script would be used in JupyterLab to launch its unit tests with whatever extra config it requires.


blink1073:
@minrk, what do you think of using this approach?

afshin:
I like the existence of a server when running JupyterLab tests, it seems like a good approach for something that is fundamentally both a client and a server application.

test/integration_test.py
```diff
@@ -1,18 +1,30 @@
...
+
 
 KARMA_PORT = 9876
 
 
 def start_notebook():
     nb_command = [sys.executable, '-m', 'notebook', '--no-browser',
-                  '--debug', '--NotebookApp.allow_origin="*"']
+                  '--debug', '--NotebookApp.allow_origin="*"',
+                  '--NotebookApp.notebook_dir="%s"' % root_dir]
```
minrk: Minor detail: notebook_dir is already a positional arg, so the `notebook_dir=` can be left out, leaving Popen to deal with quoting/escaping root_dir.

blink1073: Done.

Issue #236, opened 2016-10-06
Use the new sessions API


blink1073:
cf https://github.com/jupyter/notebook/pull/1359#issuecomment-252044121

PR #237, opened 2016-10-06
[WIP] Use the new session api

blink1073:
cf #236 

cc @minrk

### DIFF
--- a/src/session/default.ts
+++ b/src/session/default.ts
@@ -79,9 +79,9 @@ class DefaultSession implements ISession {
   unhandledMessage: ISignal<ISession, KernelMessage.IMessage>;
 
   /**
-   * A signal emitted when the session path changes.
+   * A signal emitted when the session model changes.
    */
-  pathChanged: ISignal<ISession, string>;
+  modelChanged: ISignal<ISession, Session.IModel>;
 
   /**
    * Get the session id.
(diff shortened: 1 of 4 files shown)


blink1073:
The updated integration tests pass.

src/session/default.ts
```diff
@@ -211,13 +230,45 @@ class DefaultSession implements ISession {
...
+   *
+   * #### Notes
+   * This uses the Jupyter REST API, and the response is validated.
+   * The promise is fulfilled on a valid response and rejected otherwise.
+   */
+  setType(type: string): Promise<void> {
+    if (this.isDisposed) {
+      return Promise.reject(new Error('Session is disposed'));
+    }
+    let data = JSON.stringify({ type });
```
minrk: I did not know that you could use this shorthand for objects!

blink1073: Yep, available in ES6.   I am going to also verify that current master of services still works against notebook master (that the old api still works).

blink1073:
Tests pass:

```
Session
[W 09:10:11.490 NotebookApp] Sessions API changed, see updated swagger docs
...
  18 passing (14s)
```

minrk:
Very nice.

blink1073:
Closing as stale.

PR #238, opened 2016-10-07
More API Updates

blink1073:
Update the ServiceManager API to the new style.
Rename `Content.FileType` -> `Content.ContentType` to avoid ambiguity with the `FileType` object in JupyterLab.


afshin:
👍

PR #244, opened 2016-10-11
Types used in the codebase need to be installed as normal dependencies.

jasongrout:
Otherwise the types are not installed for packages that depend on this package.


jasongrout:
We should probably bump and release jupyter-js-services 0.21.1 with this change.

afshin:
👍

PR #248, opened 2016-10-13
Avoid having an explicit node typings requirement

blink1073:
We also don't need the `minimist` `@typings` in `dependencies`, because it isn't in the compiled `.d.ts` file.

cc @jasongrout


blink1073:
Superseded by #249

PR #249, opened 2016-10-13
Typings

jasongrout:
This cleans up the typings to (almost) conform to a Typescript 2.0 philosophy.

Some problems we were facing before:

1. Because the dev dependencies included the `ws` package, we were pulling in typings for that, which in turn pulled in `@types/node`. However, that meant that for us and all the downstream packages, the node typings were active, which redefined tons of stuff. In particular, I was having issues with node redefining `setTimeout` to return something apparently other than a number, which in turn was causing issues with JupyterLab using setTimeout. Part of the reason that `@types/node` became an issue was because we are using the `url` package in the utilities file, which in turn caused a reference to the node typings to be inserted in the generated file. I solve this problem by including a `types` key in the `tsconfig.json` to narrow the library types to things we actually use.

2. In order to avoid having downstream dependencies needing to include typings files for the url package (which does not exist on `@types/url`), I just define the interface in the utility file where it is needed. This could be fixed when `@types/url` exists.

3. Typedoc still does not recognize the tsconfig's `lib` key, so we still have to manually include those files in the typings.d.ts file.


vidartf:
Just as an FYI, the `setTimeout` issue can be resolved by using `window.setTimeout`, although that might not be ideal?

blink1073:
This looks good to me pending my PR against this PR.

jasongrout:
Some issues I've only seen when building downstream packages against this package, so I was holding off on declaring this finished until I had worked out typings in jupyterlab and ipywidgets.

jasongrout:
> Just as an FYI, the setTimeout issue can be resolved by using window.setTimeout, although that might not be ideal?

Good point! Though the node typings still shouldn't have been used, so it's good we cleared up an underlying issue as well.

jasongrout:
Yet again Travis came back with:

>  1) session ISession #statusChanged should emit when the kernel status changes:
>   Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.

Restarting the test...

blink1073:
I spent a lot of time yesterday trying to hunt down those spurious errors in #250.  As best as I can tell, the issue comes from a race condition in creating sockets using `ws` that only happens "sometimes".

PR #253, opened 2016-10-15
More cleanup of types

jasongrout:
We separate out the hacks we need for typedoc, and clean up the requirejs typings and the test typings.


jasongrout:
We should also enable the requirejs typings for the test directory, but when we do, we get conflicts between the node typings that are already loaded and the requirejs typings. So we leave in the hack in `test/src/target.ts` that defines `define`.

blink1073:
LGTM, thanks!

PR #254, opened 2016-10-18
Use iterators and sequences

blink1073:
Avoid giving back raw arrays as return values.

### DIFF
--- a/src/kernel/manager.ts
+++ b/src/kernel/manager.ts
@@ -92,7 +92,7 @@ class KernelManager implements Kernel.IManager {
         this._running = running;
         this.runningChanged.emit(iter(running));
       }
-      return running;
+      return iter(running);
     });
   }
 
--- a/src/kernel/manager.ts
+++ b/src/kernel/manager.ts
@@ -2,12 +2,20 @@
 // Distributed under the terms of the Modified BSD License.
 
 import {
-  IIterator, iter, toArray
+  iter, toArray
 } from 'phosphor/lib/algorithm/iteration';
 
 import {
   deepEqual
 } from 'phosphor/lib/algorithm/json';
+
+import {
+  ISequence
+} from 'phosphor/lib/algorithm/sequence';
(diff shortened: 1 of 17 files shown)


src/kernel/manager.ts
```diff
@@ -81,13 +93,14 @@ class KernelManager implements Kernel.IManager {
...
+    return Kernel.listRunning(this._getOptions(options)).then(it => {
+      let running = toArray(it);
       if (!deepEqual(running, this._running)) {
-        this._running = running.slice();
-        this.runningChanged.emit(running);
+        this._running = running;
+        this.runningChanged.emit(new Vector(this._running));
       }
-      return running;
+      return iter(running);
```
afshin: Can return `it` instead of `iter(running)` or is there a mutability issue there?

src/session/default.ts
```diff
@@ -501,7 +505,7 @@ namespace Private {
       }
     }
     return listRunning(options).then(models => {
-      for (let model of models) {
+      for (let model of toArray(models)) {
```
afshin: Maybe `each` instead of `for`.

src/session/manager.ts
```diff
@@ -84,13 +96,14 @@ class SessionManager implements Session.IManager {
...
+    return Session.listRunning(this._getOptions(options)).then(it => {
+      let running = toArray(it);
       if (!deepEqual(running, this._running)) {
-        this._running = running.slice();
-        this.runningChanged.emit(running);
+        this._running = running;
+        this.runningChanged.emit(new Vector(running));
       }
-      return running;
+      return iter(running);
```
afshin: Same comment about returning `it` vs. `iter(running)`.

src/kernel/manager.ts
blink1073: We've already exhausted the iterator in `toArray`

afshin:
👍

PR #263, opened 2016-10-28
API Cleanup and Manager State Management

### LINKED ISSUE #264: Cache the kernel specs and look them up by name
This avoids having multiple trips to the server for information we already have and allows us to work around the error seen in https://github.com/jupyter/notebook/issues/1859 without waiting for a new release.

blink1073:
- Adds `terminated` signals to terminals, kernels, and sessions.
- Adds `specs` and `running()` conveniences to the terminal, kernel, and session managers.
- The managers now maintain the state of the running models when there is a known change or on a polling interval.
- Update the message request method names to be of the form `requestFoo`.
- Normalizes the api of kernel, terminal, and session and their managers.
- Fixes #264.

These changes will alleviate the need to call `listRunning()` manually from different places and rely on a signal handler to get the most recent running models.


src/terminal/default.ts
```diff
@@ -0,0 +1,358 @@
...
+   * otherwise the promise is rejected.
+   *
+   * If the session was not already started and no `options` are given,
+   * the promise is rejected.
+   */
+  export
+  function connectTo(name: string, options: TerminalSession.IOptions = {}): Promise<TerminalSession.ISession> {
+    options.baseUrl = options.baseUrl || utils.getBaseUrl();
+    let url = Private.getTermUrl(options.baseUrl, name);
+    if (url in Private.running) {
```
afshin: You might want to consider using a `Set` for `running`, and checking with the `has()` method since that's a pattern we're starting to use elsewhere.

src/terminal/default.ts
```diff
@@ -0,0 +1,358 @@
...
+      }
+      let data = success.data as TerminalSession.IModel[];
+      if (!Array.isArray(data)) {
+        return utils.makeAjaxError(success, 'Invalid terminal data');
+      }
+      // Update the local data store.
+      let urls = toArray(map(data, item => {
+          return utils.urlPathJoin(url, item.name);
+      }));
+      each(Object.keys(Private.running), runningUrl => {
```
afshin: *e.g.* if `Private.running` were a `Set` you could do `Private.running.forEach(...)`

afshin:
👍 

Awesome, thanks!

PR #265, opened 2016-11-01
Update the specs interface

blink1073:
This makes it easier to consume the initial specs if that is all that is desired.


afshin:
👍

PR #267, opened 2016-11-02
Handle object readiness and clean up manager auto-refresh

blink1073:
- Adds `ready(): Promise<void>;` functions to all objects that require data from the server before being fully initialized.
- Cleans up auto-refresh behavior of the managers.  They all refresh their listings at 10 second intervals, and kernelspecs are fetched at 61 second intervals (to avoid overlap).
- This auto-refresh behavior should eventually be replaced by a WebSocket connection to the server.
- Verified against JLab using `npm link`.

### DIFF
--- a/src/kernel/manager.ts
+++ b/src/kernel/manager.ts
@@ -79,8 +79,8 @@ class KernelManager implements Kernel.IManager {
       return;
     }
     this._isDisposed = true;
-    clearTimeout(this._runningTimer);
-    clearTimeout(this._specsTimer);
+    clearInterval(this._runningTimer);
+    clearInterval(this._specsTimer);
     clearSignalData(this);
     this._specs = null;
     this._running = [];
--- a/src/kernel/manager.ts
+++ b/src/kernel/manager.ts
@@ -228,7 +228,6 @@ class KernelManager implements Kernel.IManager {
         this._running = running.slice();
         this.runningChanged.emit(running);
       }
-      return running;
     });
   }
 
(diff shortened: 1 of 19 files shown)


jasongrout:
> should eventually be replaced by a WebSocket connection to the server

Another option is [Server Sent Events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events), but I think we'll need a websocket connection with the server anyway to support collaboration.

src/kernel/manager.ts
```diff
@@ -67,8 +79,8 @@ class KernelManager implements Kernel.IManager {
       return;
     }
     this._isDisposed = true;
-    clearTimeout(this._updateTimer);
-    clearTimeout(this._refreshTimer);
+    clearTimeout(this._runningTimer);
+    clearTimeout(this._specsTimer);
```
afshin: Technically, these calls will work: https://developer.mozilla.org/en-US/docs/Web/API/WindowTimers/clearTimeout

But because these are intervals, it's better to use `clearInterval`.

src/session/manager.ts
```diff
@@ -71,8 +83,8 @@ class SessionManager implements Session.IManager {
       return;
     }
     this._isDisposed = true;
-    clearTimeout(this._updateTimer);
-    clearTimeout(this._refreshTimer);
+    clearTimeout(this._runningTimer);
+    clearTimeout(this._specsTimer);
```
afshin: Ditto re: `clearInterval`

PR #273, opened 2016-11-07
Add isReady attributes and other cleanup

### LINKED ISSUE #271: Make the kernel spec on a given kernel a promise
Move it out of the "ready" and make it a separate promise.  There is no need to store a cached version.

### LINKED ISSUE #272: Add "isReady" flags to all objects that have "ready()" methods
Allows consumers to avoid the promise if the object is already ready.

blink1073:
- Adds `isReady` attribute to objects that have `ready()` promises.
- Moves `kernel.spec` -> `kernel.spec(): Promise` and removes it from a kernel's readiness
- Cleanup and refactoring of tests
- Fix bug in terminal session internal handling

Fixes #271.  Fixes #272.


src/terminal/default.ts
```diff
@@ -275,7 +291,7 @@ namespace DefaultTerminalSession {
           return utils.urlPathJoin(url, item.name);
       }));
       each(Object.keys(Private.running), runningUrl => {
-        if (urls.indexOf(runningUrl) !== -1) {
+        if (urls.indexOf(runningUrl) === -1) {
```
afshin: Nice. One char fix.

PR #276, opened 2016-11-15
Add more intelligence to the managers

blink1073:
Updates the running session/kernel state when client state changes, avoiding an out of band server request.  It can potentially yield false positives for a kernel or session terminating, but will get back in sync after the next refresh cycle.


afshin:
👍

PR #281, opened 2016-11-17
add token option

### LINKED ISSUE #251: Examples of connecting to secured servers
Hi, I'm working on remote kernel functionality in [nteract/hydrogen](https://github.com/nteract/hydrogen), using this services package.

Lately we've had several reports of being unable to configure secure connections (see https://github.com/nteract/hydrogen/issues/467 and https://github.com/nteract/hydrogen/issues/473). I'm trying to address these issues, but there doesn't appear to be any documentation on how to connect `jupyter-js-services` to a secured server (be it a notebook server or a kernel gateway).

Could someone involved with `jupyter-js-services` or the Jupyter project help me out here?

Based on crawling a scattering of github issues, websites, and mailing list posts, my current understanding of the situation is:
- Notebooks don't appear to support any auth mechanism other than cookies, meaning that programs have no chance of connecting to secure notebook servers unless they embed a web browser
- Kernel gateways support token-based auth, but the setup required is not documented anywhere. My best attempt was setting up `KG_AUTH_TOKEN` and passing `requestHeaders.Authorization`, which allows accessing the HTTP-based API. However attempting to actually connect to a kernel over websockets does not pass the required credentials, and I haven't come across any documentation on how to get the websocket connection to proceed.
- https://github.com/jupyterlab/services/pull/206 only appears to address the use case of frontends running in the browser. If you don't have a browser-based cookie system available, the method there doesn't work.

It is my belief that supporting community-made frontends is beneficial for the Jupyter community, and the kernel gateway + jupyter-js-services combination provides a good foundation for this. However, I think that being able to establish secure connections is an important part of this story. I'd appreciate if someone could help me get this feature working.

Thanks!

minrk:
added to ajaxSettings.requestHeaders['Authorization'] for ajax requests, and url parameter for websockets.

This should work with Kernel Gateway now, and notebook >= 4.3. It should be safely ignored on notebook < 4.3.

closes #251

### DIFF
--- a/examples/browser-require/main.py
+++ b/examples/browser-require/main.py
@@ -24,12 +24,13 @@ class MainPageHandler(tornado.web.RequestHandler):
 
 def main():
     # Start a notebook server with cross-origin access.
-    url = "http://localhost:%s" % PORT
-
-    nb_command = [sys.executable, '-m', 'notebook','--no-browser', '--debug',
-                  '--NotebookApp.password=""',
-                  '--notebook-dir=.',
-                  '--NotebookApp.allow_origin="%s"' % url]
+    nb_command = [sys.executable, '-m', 'notebook', '--no-browser',
+                  '--debug',
+                  # FIXME: allow-origin=* only required for notebook < 4.3
+                  '--NotebookApp.allow_origin="*"',
+                  # disable user password:
+                  '--NotebookApp.password=',
+              ]
     nb_server = subprocess.Popen(nb_command, stderr=subprocess.STDOUT,
                                  stdout=subprocess.PIPE)
 
--- a/examples/browser/main.py
+++ b/examples/browser/main.py
@@ -24,10 +24,13 @@ class MainPageHandler(tornado.web.RequestHandler):
 
 def main():
     # Start a notebook server with cross-origin access.
-    url = "http://localhost:%s" % PORT
-
-    nb_command = [sys.executable, '-m', 'notebook', '--no-browser', '--debug',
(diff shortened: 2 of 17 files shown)


blink1073:
Why not add `token` to `IAjaxSettings` and let it be applied in `ajaxRequest` if given?

minrk:
@blink1073 I can do that, too, if that sounds better to you. I wasn't sure if that was best, since it's not strictly ajax (it applies to websockets as well). I also thought it might be appropriate at the top-level, since ~all nodejs uses of services will need it going forward, as auth becomes required by default.

minrk:
I got the tests passing (I think). So then the question should become balancing user convenience of a top-level `token` option vs the internal simplicity of storing the token in ajaxSettings. The `ajaxSettingsWithToken` everywhere isn't my favorite thing, so I'm fine removing it, if you prefer.

I don't understand the failures on 0.12.

blink1073:
I think what you have here is fine.  It is conceptually simpler, just messier in implementation.

The current error looks legitimate, I'm looking into it now:

```python
Traceback (most recent call last):
  File "main.py", line 74, in <module>
    main()
  File "main.py", line 30, in main
    base_url = re.search('(http.*?)[\?$]', line).groups()[0]
AttributeError: 'NoneType' object has no attribute 'groups'
```

minrk:
Thanks! If you have a good idea for centralizing the token so the merging into ajaxSettings doesn't have to be so repetitive, that would be great. For now, this appears to work.

blink1073:
We can roll with this, thanks!

PR #283, opened 2016-11-20
Handle terminal availability

blink1073:
cf https://github.com/jupyterlab/jupyterlab/issues/1265


afshin:
Awesome. The best defense is a good defense.

PR #291, opened 2017-01-07
Compatibility with Notebook 4.3.1

blink1073:
Tested against Notebook 4.3.1 and JupyterLab.

- Makes sure all requests have a body. 
- The `xsrf` cookie was not working properly when using Firefox on Karma, so we add a check for the existence of a token auth before using the cookie.
- Cleans up the python scripts to use a notebook app directly.

### DIFF
--- a/examples/browser-require/index.html
+++ b/examples/browser-require/index.html
@@ -33,6 +33,6 @@
         }
     });
     </script>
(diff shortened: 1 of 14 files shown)


afshin:
👍

minrk:
Nice!

PR #297, opened 2017-01-14
Use promise for getDownloadUrl

ian-r-rose:
In some cases the downloadUrl for a file may not be computable client-side and requires some server communication. Cf. jupyterlab/jupyterlab#1495 and [https://github.com/ian-r-rose/jupyterlab_google_drive/tree/filebrowser]()


blink1073:
Nice, thanks!

PR #302, opened 2017-01-30
Make sure the internal state is updated before emitting

blink1073:
We were emitting a path changed before setting the internal path value, meaning `session.path` and `session.model` not updated in `pathChanged` signal handlers.


blink1073:
cf https://github.com/jupyterlab/jupyterlab/issues/1570

PR #310, opened 2017-03-09
Remove url dependencies

blink1073:
[url](https://github.com/defunctzombie/node-url) hasn't been maintained and depends on the deprecated version of `punycode` that gets overwritten by a new version when using Webpack 2.0.


src/kernel/default.ts
```diff
@@ -655,7 +657,6 @@ class DefaultKernel implements Kernel.IKernel {
     if (this._token !== '') {
       url = url + `&token=${encodeURIComponent(this._token)}`;
```
afshin: Is there always a guaranteed query string by this point in the logic? Because if not, this should probably be:
```typescript
url += `${url.search ? '&' : '?'}token=${encodeURIComponent(this._token)}`;
```

.travis.yml
```diff
@@ -13,5 +13,5 @@ install:
 - bash ./scripts/travis_install.sh
 script:
 - bash ./scripts/travis_script.sh
-after_success:
-- bash ./scripts/travis_after_success.sh
+# after_success:
+# - bash ./scripts/travis_after_success.sh
```
afshin: Do these need to remain commented out?

scripts/travis_script.sh
```diff
@@ -15,4 +15,4 @@ pushd examples/node
 python main.py 
 popd
 
-npm run docs
+#npm run docs
```
afshin: Also these?

src/kernel/default.ts
```diff
@@ -644,8 +644,10 @@ class DefaultKernel implements Kernel.IKernel {
     let partialUrl = utils.urlPathJoin(this._wsUrl, KERNEL_SERVICE_URL,
                                        encodeURIComponent(this._id));
     // Strip any authentication from the display string.
-    let parsed = utils.urlParse(partialUrl);
-    let display = partialUrl.replace(parsed.auth, '');
+    if (typeof document !== 'undefined') {
```
afshin: Since this check is already done in the parse function, instead of relying on the client to *know* how it was implemented, maybe it should instead be:
```typescript
try {
  let parsed = utils.urlParse(partialUrl);
  console.log('Starting websocket', parsed.hostname);
} catch (error) { /* Ignore error. */ }
```

src/utils.ts
```diff
@@ -79,62 +76,59 @@ function uuid(): string {
...
- *   to an object returned by the `querystring` module's `parse()` method.
- *   If `false`, the `query` property on the returned URL object will be an
- *   unparsed, undecoded string. Defaults to `false`.
- *
- * @param slashedDenoteHost - If `true`, the first token after the literal
- *   string `//` and preceeding the next `/` will be interpreted as the `host`.
- *   For instance, given `//foo/bar`, the result would be
- *   `{host: 'foo', pathname: '/bar'}` rather than `{pathname: '//foo/bar'}`.
- *   Defaults to `false`.
+ * @param url - The URL string to parse.
```
afshin: This should have an `@throws` line in the documentation.

.travis.yml
blink1073: Yes, TypeDoc no longer works for this repo.

src/kernel/default.ts
blink1073: It's already there.

src/utils.ts
blink1073: No longer throws.

.travis.yml
blink1073: Fixed by pinning to TS 2.1.

PR #312, opened 2017-03-14
Limit the kernel interface in a session

blink1073:
Do not expose the signals or the shutdown method, to force them to be used on the session itself.

cf https://github.com/jupyterlab/jupyterlab/issues/1856


blink1073:
Absorbed into https://github.com/jupyterlab/jupyterlab/pull/1927.

PR #314, opened 2017-03-17
Terminal session disposed checks.

afshin:


blink1073:
Apparently we can't run the docs on Node v0.12 anymore.

